PR #3, opened 2024-04-16
fix: copy `polkavm_guest.bc` to output directory

0xf333:
Copying the generated `polkavm_guest.bc` file to the out put directory specified by the user.

### E.g 
The following command
 ````bash
 cargo run Storage.sol -O3 --bin --output-dir './out'
 ```` 
 Should additionally output `polkavm_guest.bc` to `./out` directory.
 
 ---
 
 ### Current behavior
 In this draft, I'm temporary copying it to `current` directory, I will implement the fix before the PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /target
 *.dot
+*.bc
 .vscode/
 .DS_Store
 /*.sol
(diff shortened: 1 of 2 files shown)


0xf333:
There won't be a need for this after all, let close this daft.

PR #5, opened 2024-04-18
feat: streamline bin updates during developement

0xf333:
## Description
This PR introduces a way for the `zksolc` binary to automatically be updated/copied to `.cargo/bin` after each build to resolves the issue of having to manually update the $PATH or each time keep running `cargo install --path crates/solidity`, which was a recurring annoyance during testing and development.

## Usage

To install the project dependencies, run:
```bash
make install
```

To clean the project directory from all build artifacts and dependencies, run:
```bash
make clean
```
To run integration tests, use
```bash
make test-integration
```

To run the project's tests, ( both Rust and npm tests ) use:
```bash
make test
```

To run only solidity crate test, use:
```bash
make test-solidity
```

To run only the CLI tests, use:
```bash
make test-cli
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,24 +1,23 @@
-.PHONY : install test test-solidity test-cli test-integration clean
+.PHONY: install test test-solidity test-cli test-integration clean
 
-install :
-        cargo install --path crates/solidity && \
-        npm install && npm fund
+install:
+	cargo install --path crates/solidity && \
+	npm install && npm fund
 
-test : install test-integration test-cli test-solidity
+test: install test-integration test-cli test-solidity
 
-test-integration : install
-        cargo test --package revive-integration
+test-integration: install
+	cargo test --package revive-integration
 
-test-solidity : install
-        cargo test --package revive-solidity
+test-solidity: install
+	cargo test --package revive-solidity
 
-test-cli : install
-        npm run test:cli
+test-cli: install
+	npm run test:cli
 
-clean :
-        cargo clean && \
-        rm -rf node_modules && \
-        rm -rf crates/solidity/src/tests/cli-tests/artifacts && \
-        rm -f ~/.cargo/bin/zksolc && \
-        rm -f package-lock.json
- 
+clean:
+	cargo clean && \
+	rm -rf node_modules && \
+	rm -rf crates/solidity/src/tests/cli-tests/artifacts && \
+	rm -f ~/.cargo/bin/zksolc && \
+	rm -f package-lock.json
--- a/Makefile
+++ b/Makefile
@@ -1,24 +1,24 @@
-install:
-	cargo install --path crates/solidity && \
-	npm install && npm fund
+.PHONY : install test test-solidity test-cli test-integration clean
 
-test:
-	cargo install --path crates/solidity && \
-	npm install && npm fund && \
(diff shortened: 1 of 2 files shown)


0xf333:
@xermicus

xermicus (commented):
Thanks for the PR! I like the idea but to me it'd make much more sense to have:
- A `install`, `test` `test-solidity` and `test-cli` target
- `test` installs and then invokes all tests
- `test-solidity` and `test-cli` installs and calls the according tests

WDYT?

Makefile
```diff
@@ -0,0 +1,19 @@
+build:
+	cargo build && \
+	cp target/debug/zksolc ~/.cargo/bin/zksolc && \
+	npm install && npm fund
+
+build-release:
+	cargo build --release && \
```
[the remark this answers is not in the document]
0xf333: Ok, making the change now

0xf333:
> WDYT?

Yep, that would be more simple, I'm making the change now

Makefile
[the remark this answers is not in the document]
0xf333: @xermicus 
suggested change implemented

Makefile
```diff
@@ -0,0 +1,24 @@
...
+	npm run test:cli
+
+clean:
+	cargo clean && \
+	rm -rf node_modules && \
+	rm -rf crates/solidity/src/tests/cli-tests/artifacts && \
+	rm -f ~/.cargo/bin/zksolc && \
+	rm -f package-lock.json
+
+
```
xermicus, line 24:
```suggestion
.PHONY : install test test-solidity test-cli test-integration clean

install :
        cargo install --path crates/solidity && \
        npm install && npm fund

test : install test-integration test-cli test-solidity

test-integration : install
        cargo test --package revive-integration

test-solidity : install
        cargo test --package revive-solidity

test-cli : install
        npm run test:cli

clean :
        cargo clean && \
        rm -rf node_modules && \
        rm -rf crates/solidity/src/tests/cli-tests/artifacts && \
        rm -f ~/.cargo/bin/zksolc && \
        rm -f package-lock.json
 
```

We should call the targets directly this way it's much more maintainable :) Also we'll add more in the future (more tests. I'm about to integrate a benchmarking suite etc..) and I think it'd make sense to use the Makefile for those as well :)

xermicus (changes requested):

Makefile
0xf333: @xermicus 
Commit add to the PR

Makefile
```diff
@@ -0,0 +1,23 @@
...
+	cargo test --package revive-solidity
+
+test-cli: install
+	npm run test:cli
+
+clean:
+	cargo clean && \
+	rm -rf node_modules && \
+	rm -rf crates/solidity/src/tests/cli-tests/artifacts && \
+	rm -f ~/.cargo/bin/zksolc && \
```
xermicus, line 22: `$HOME/.cargo/` is just the default, we should use `cargo uninstall instead`

```suggestion
	cargo uninstall revive-solidity
```

xermicus (commented):
Just have a last small nit, otherwise LGTM!

0xf333:
> Just have a last small nit, otherwise LGTM!

commit pushed

xermicus (approved):
Thanks!

PR #6, opened 2024-04-23
feat: use PolkaVM disassembler

0xf333:
## Desciption
  - Integratd the disassembler
  - updated `asm.test` to check for the correct patterns in the assembly text.
  - addressed tsconfig conflict

### Usage example 
```bash
zksolc contract.sol --asm --output-dir './tmpdir'
```
The assembly `.zasm` file will be outputed in the `tmpdir` directory

---

### `--asm` flag test
```bash
make test-cli
```
Now `asm.test` working properly

### DIFF
--- a/crates/solidity/src/zksolc/main.rs
+++ b/crates/solidity/src/zksolc/main.rs
@@ -6,6 +6,9 @@ use std::str::FromStr;
 
 use self::arguments::Arguments;
 
+use polkavm_common::program::ProgramBlob;
+use polkavm_disassembler::{Disassembler, DisassemblyFormat};
+
 /// The rayon worker stack size.
 const RAYON_WORKER_STACK_SIZE: usize = 16 * 1024 * 1024;
 
@@ -190,18 +193,28 @@ fn main_inner() -> anyhow::Result<()> {
         );
     } else if arguments.output_assembly || arguments.output_binary {
         for (path, contract) in build.contracts.into_iter() {
+            let bytescode = contract.build.bytecode;
+
             if arguments.output_assembly {
-                println!(
-                    "Contract `{}` assembly:\n\n{}",
-                    path, contract.build.assembly_text
-                );
+                let program_blob = match ProgramBlob::parse(bytescode.as_slice()) {
+                    Ok(blob) => blob,
+                    Err(error) => {
+                        anyhow::bail!("Failed to parse program blob: {}", error);
+                    }
+                };
+
+                let disassembler_object =
+                    Disassembler::new(&program_blob, DisassemblyFormat::Guest)?;
+
+                let mut disassembled_code = Vec::new();
+                disassembler_object.disassemble_into(&mut disassembled_code)?;
+
+                let assembly_text = String::from_utf8(disassembled_code)?;
+
+                println!("Contract `{}` assembly:\n\n{}", path, assembly_text);
             }
             if arguments.output_binary {
-                println!(
-                    "Contract `{}` bytecode: 0x{}",
-                    path,
-                    hex::encode(contract.build.bytecode)
-                );
+                println!("Contract `{}` bytecode: 0x{}", path, hex::encode(bytescode));
             }
         }
     } else {
--- a/crates/solidity/src/zksolc/main.rs
+++ b/crates/solidity/src/zksolc/main.rs
@@ -6,6 +6,7 @@ use std::str::FromStr;
 
 use self::arguments::Arguments;
 
+use anyhow::{bail, Context};
 use polkavm_common::program::ProgramBlob;
 use polkavm_disassembler::{Disassembler, DisassemblyFormat};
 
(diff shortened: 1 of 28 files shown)


0xf333:
### Introduced change
  - Integrated the disassembler
  - updated `.gitginore` 

### Usage example 
```bash
zksolc contract.sol --asm --output-dir './tmpdir'
```
The assembly `.zasm` file will be outputed in the `tmpdir` directory

0xf333:
### Introduced change
   - Now asm print in `main_inner` to output the correct assembly text.
   - Now `asm.test` to check for the correct patterns in the assembly text.
   
### Usage
```bash
make test-cli
```
Now `asm.test` working properly

0xf333:
@xermicus

crates/solidity/src/zksolc/main.rs
```diff
@@ -190,18 +194,36 @@ fn main_inner() -> anyhow::Result<()> {
...
-                println!(
-                    "Contract `{}` assembly:\n\n{}",
-                    path, contract.build.assembly_text
-                );
+                let program_blob = match ProgramBlob::parse(bytescode.as_slice()) {
+                    Ok(blob) => blob,
+                    Err(error) => {
+                        bail!("Failed to parse program blob: {}", error);
+                    }
+                };
```
xermicus, line 205: Total nitpick. But you can do instead:

```suggestion
                let program_blob = ProgramBlob::parse(bytescode.as_slice())
                    .map_err(|error| anyhow::anyhow!("Failed to parse program blob: {error}"))?;
```

More concise and I personally find it easier to read. But it's a question of style and personal preference so I'll leave it up to you.

crates/solidity/src/zksolc/main.rs
```diff
@@ -190,18 +194,36 @@ fn main_inner() -> anyhow::Result<()> {
...
+
+                let disassembler_object =
```
xermicus, line 207: Another nitpick, in rust land we would name this variable just `disassembler` :upside_down_face:

crates/solidity/src/zksolc/main.rs
```diff
@@ -190,18 +194,36 @@ fn main_inner() -> anyhow::Result<()> {
...
+                    Disassembler::new(&program_blob, DisassemblyFormat::Guest).context(format!(
+                        "Failed to create disassembler for contract '{}'",
+                        path
+                    ))?;
+
```
xermicus, line 212: I think we could instruct the disassembler to emit gas usage (`disassembler.display_gas()?;`).

xermicus (commented):
Thanks!

I think since we now have an assembly text format, [`fn build_assembly_text`](https://github.com/0xf333/revive/blob/0xf333/addeding_revivetool_features/crates/llvm-context/src/polkavm/mod.rs#L23) should be fixed too

crates/solidity/src/zksolc/main.rs
0xf333: yep, I will add it now

crates/solidity/src/zksolc/main.rs
0xf333: ok

0xf333:
> Thanks!
> 
> I think since we now have an assembly text format, [`fn build_assembly_text`](https://github.com/0xf333/revive/blob/0xf333/addeding_revivetool_features/crates/llvm-context/src/polkavm/mod.rs#L23) should be fixed too

Yes, I actually planned `build_assembly_text` refactoring to go along with the `--zkasm` fix, which will be in my next pull request.
There seems to be more to it than simply giving it the correct assembly text.

crates/solidity/src/zksolc/main.rs
```diff
@@ -190,18 +193,41 @@ fn main_inner() -> anyhow::Result<()> {
...
+                        ))
+                    })?;
+
+                let assembly_text = String::from_utf8(disassembled_code)
+                    .map_err(|error| anyhow::anyhow!(format!(
+                        "Failed to convert disassembled code to string for contract '{}'\n\nDue to:\n{}",
+                        path, error
+                    )))?;
+
+                println!("Contract `{}` assembly:\n\n{}", path, assembly_text);
```
xermicus, line 227: Isn't this redudant now since we construct the disassembly in `crates/solidity/src/build/contract.rs`?

xermicus:
> > Thanks!
> > I think since we now have an assembly text format, [`fn build_assembly_text`](https://github.com/0xf333/revive/blob/0xf333/addeding_revivetool_features/crates/llvm-context/src/polkavm/mod.rs#L23) should be fixed too
> 
> Yes, I actually planned `build_assembly_text` refactoring to go along with the `--zkasm` fix, which will be in my next pull request. There seems to be more to it than simply giving it the correct assembly text.

Right, so this will remove the redundant parts and just use `contract.build` everywhere

0xf333:
> Right, so this will remove the redundant parts and just use `contract.build` everywhere

Yeah, that's what I'm going for.

PR #8, opened 2024-05-03
chore: added code formating target

0xf333:
## Description
Since 'make test' is frequently used during development, adding the codebase formatting target as a prerequisite to it, will help maintain consistency in the codebase more easily.

### Usage
During testing, formatting will automatically be applied 
```bash
make test
```
#### Or

If you only want to format the codebase without running the tests, you should just run:
```bash
make format
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ install-npm:
 	npm install && npm fund
 
 format:
-	cargo fmt --all
+	cargo fmt --all --check
 
 test: format install test-integration test-cli test-solidity
 
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: install test test-solidity test-cli test-integration clean
+.PHONY: install format test test-solidity test-cli test-integration clean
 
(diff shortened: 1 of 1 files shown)


0xf333:
@xermicus

xermicus (commented):
Good idea however can we just `--check` instead of applying? This way executing tests won't potentially change source code (which is a very unexpected thing to do for tests). Also I'm about to add basic CI jobs where we want to re-use the Makefile targets, and with `--check` it'll be directly reusable since it returns non-zero exit code if it finds unfromatted code.

0xf333:
> Good idea however can we just `--check` instead of applying? This way executing tests won't potentially change source code (which is a very unexpected thing to do for tests). Also I'm about to add basic CI jobs where we want to re-use the Makefile targets, and with `--check` it'll be directly reusable since it returns non-zero exit code if it finds unformatted code.

@xermicus 
I see what you mean. 
Commit pushed with the suggested change

xermicus (approved):
Thanks!

PR #9, opened 2024-05-07
fix: addressing assembly text build mechanism

0xf333:
Description
===========
  - Centralized build mechanism to make the implementation easy to maintain


0xf333:
@xermicus

crates/llvm-context/src/polkavm/mod.rs
```diff
@@ -22,64 +25,51 @@ pub fn initialize_target() {
...
+        .map_err(|error| anyhow::anyhow!(format!("Failed to parse program blob:\n{}\n", error)))?;
+
+    let mut disassembler =
+        Disassembler::new(&program_blob, DisassemblyFormat::Guest).map_err(|error| {
+            anyhow::anyhow!(format!(
+                "Failed to create disassembler for contract:\n{:?}\n\nDue to:\n{}\n",
+                contract_path, error
+            ))
+        })?;
+
```
xermicus, line 49:
```suggestion
    disassembler.display_gas()?;
```

We always want to display the gas usage, not only on error :)

crates/llvm-context/src/polkavm/mod.rs
```diff
@@ -22,64 +25,51 @@ pub fn initialize_target() {
...
+    let mut disassembled_code = Vec::new();
+    disassembler
+        .disassemble_into(&mut disassembled_code)
         .map_err(|error| {
```
xermicus, line 53:
```suggestion
        .context("Failed to disassemble contract: {contract_path}")?;
```
I really like providing useful information with the error, however (nitpick incoming :sweat_smile: ) I think the idiomatic way would be to use an error context here.

crates/llvm-context/src/polkavm/mod.rs
```diff
@@ -22,64 +25,51 @@ pub fn initialize_target() {
...
-        }
-    }
-    .map(hex::encode)
-    .map_err(|_error| {
-        anyhow::anyhow!("The contract `{}` bytecode hashing error", contract_path,)
+    let assembly_text = String::from_utf8(disassembled_code).map_err(|error| {
+        anyhow::anyhow!(format!(
+            "Failed to convert disassembled code to string for contract\n{:?}\n\nDue to:\n{}\n",
+            contract_path, error
+        ))
```
xermicus, line 66:
```suggestion
    let assembly_text = String::from_utf8(disassembled_code)
        .context("Failed to convert disassembled code to string for contract: {contract_path}")?;
```

Same nitpick here 

Note that for this to work you'll have to import the trait like `use anyhow::Context as AnyhowContext` or something like this, as we already import the `Context` from this crate (aliasing the anyhow context like this neatly resolves the conflict).

xermicus:
Just a few minor things, otherwise LGTM!

crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -30,6 +30,7 @@ use crate::optimizer::Optimizer;
 use crate::polkavm::r#const::*;
 use crate::polkavm::DebugConfig;
 use crate::polkavm::Dependency;
+//use crate::polkavm_evm_event::log;
```
xermicus, line 33:
```suggestion
```

I think we can as well remove it or is it related to something needing care in a followup?

0xf333: > or is it related to something needing care in a followup?

It can be removed

0xf333:
> Just a few minor things, otherwise LGTM!

Ok, commit pushed

xermicus (approved):
Thanks!

xermicus (commented):
Thanks!

PR #10, opened 2024-05-22
feat: heap GEP to allocate memory if needed

0xf333:
work in progress


0xf333:
implemented in a different commit in another PR

PR #26, opened 2024-08-02
PoC - revive wasm version

smiasojed:
Do not merge, work in progress

PR #29, opened 2024-08-12
Add Javascript test setup

wpt967:
Add a basic set up for running revive-wasm tests. This is intended as an addition the the revive on wasm work being done in smiasojed/revive.git.


xermicus:
@wpt967 what is the exact status of this PR? We should avoid letting it rot for too long and try getting it merged soon. After #66, we can investigate if the in-browser memory limitations are still bad enough. But it doesn't block this work from being finished. Can we bring this PR to a state where the compiler works fine in node where we there are basically no resource constraint? There was also #26. IIRC the only thing missing there was proper testing on the CI.

wpt967:
Defunct, closed.

PR #32, opened 2024-08-23
[solidity,llvm-context] Improve support for debugging the compiler

wpt967:
Add option --recursive-process-input <filename> for use with --recursive-process to specify the name of a file to use instead of reading from stdin.

If --debug-output-dir is set, dump the file passed to the recursive invocation of the compiler as a JSON file suitable for use with --recursive-process-input.

These changes are intended to support debugging the compiler and are only available with DEBUG builds.

### DIFF
--- a/crates/solidity/src/process/mod.rs
+++ b/crates/solidity/src/process/mod.rs
@@ -17,13 +17,27 @@ use self::output::Output;
 pub static EXECUTABLE: OnceCell<PathBuf> = OnceCell::new();
 
 /// Read input from `stdin`, compile a contract, and write the output to `stdout`.
-pub fn run() -> anyhow::Result<()> {
+pub fn run(input_file: Option<&mut std::fs::File>) -> anyhow::Result<()> {
     let mut stdin = std::io::stdin();
     let mut stdout = std::io::stdout();
     let mut stderr = std::io::stderr();
 
     let mut buffer = Vec::with_capacity(16384);
-    stdin.read_to_end(&mut buffer).expect("Stdin reading error");
+    match input_file {
+        Some(ins) => {
+            if let Err(error) = ins.read_to_end(&mut buffer) {
+                anyhow::bail!("Failed to read recursive process input file: {:?}", error);
+            }
+        }
+        None => {
+            if let Err(error) = stdin.read_to_end(&mut buffer) {
+                anyhow::bail!(
+                    "Failed to read recursive process input from stdin: {:?}",
+                    error
+                )
+            }
+        }
+    }
 
     let input: Input = revive_common::deserialize_from_slice(buffer.as_slice())?;
     if input.enable_test_encoding {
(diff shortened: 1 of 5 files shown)


crates/solidity/src/process/mod.rs
```diff
@@ -17,13 +17,21 @@ use self::output::Output;
...
 
     let mut buffer = Vec::with_capacity(16384);
-    stdin.read_to_end(&mut buffer).expect("Stdin reading error");
+    match input_file {
+        Some(ins) => {
+            ins.read_to_end(&mut buffer)
+                .expect("Error reading from input file");
+        }
+        None => {
+            stdin.read_to_end(&mut buffer).expect("Stdin reading error");
```
xermicus, line 32: Same as above :)

crates/solidity/src/process/mod.rs
```diff
@@ -17,13 +17,21 @@ use self::output::Output;
...
     let mut stdout = std::io::stdout();
     let mut stderr = std::io::stderr();
+    let mut stdin = std::io::stdin();
 
     let mut buffer = Vec::with_capacity(16384);
-    stdin.read_to_end(&mut buffer).expect("Stdin reading error");
+    match input_file {
+        Some(ins) => {
+            ins.read_to_end(&mut buffer)
+                .expect("Error reading from input file");
```
xermicus, line 29: Nit: We usually unwrap results in case we are guaranteed to never panic (the rust compiler can not now) and use `expect(...)` to explain the reason as to why the unwrap would never panic.

But since the this function does already return a `anhow::Result` it seems like here we can be just bubble up the error instead?

crates/solidity/src/process/mod.rs
```diff
@@ -74,6 +82,11 @@ pub fn call(input: Input) -> anyhow::Result<Output> {
         anyhow::anyhow!("{:?} subprocess spawning error: {:?}", executable, error)
     })?;
 
+    #[cfg(debug_assertions)]
+    if let Some(dbg_config) = &input.debug_config {
+        let _ = dbg_config.dump_stage_output(&input.contract.path, Some("stage"), &input_json);
```
xermicus, line 87: Is silently swallowing the error here intended? Intuitively should either bubble up or panic here?

xermicus (commented):
LGTM, just some nitpicks

crates/solidity/src/process/mod.rs
```diff
@@ -74,6 +93,19 @@ pub fn call(input: Input) -> anyhow::Result<Output> {
...
+    if let Some(dbg_config) = &input.debug_config {
+        let _ = dbg_config
+            .dump_stage_output(&input.contract.path, Some("stage"), &input_json)
+            .map_err(|error| -> anyhow::Result<Output> {
+                anyhow::bail!(
+                    "{:?} failed to log the recursive process output: {:?}",
+                    executable,
+                    error,
+                )
+            });
```
xermicus, line 106:
```suggestion
            })?;
```

crates/solidity/src/process/mod.rs
```diff
@@ -17,13 +17,32 @@ use self::output::Output;
...
 
     let mut buffer = Vec::with_capacity(16384);
-    stdin.read_to_end(&mut buffer).expect("Stdin reading error");
+    match input_file {
+        Some(ins) => {
+            let _ = ins
+                .read_to_end(&mut buffer)
+                .map_err(|error| -> anyhow::Result<()> {
+                    anyhow::bail!("Failed to read recursive process input file: {:?}", error);
+                });
```
xermicus, line 32:
```suggestion
                })?;
```

crates/solidity/src/process/mod.rs
```diff
@@ -17,13 +17,32 @@ use self::output::Output;
...
+        }
+        None => {
+            let _ = stdin
+                .read_to_end(&mut buffer)
+                .map_err(|error| -> anyhow::Result<()> {
+                    anyhow::bail!(
+                        "Failed to read recursive process input from stdin: {:?}",
+                        error
+                    );
+                });
```
xermicus, line 43:
```suggestion
                })?;
```

PR #38, opened 2024-08-29
Add support for --license arg

smiasojed:
Add support for --license arg required by Remix IDE
Upload `resolc` artifact

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -44,3 +44,12 @@ jobs:
 
       - name: Test CLI
         run: make test-cli
+
+      - name: Build resolc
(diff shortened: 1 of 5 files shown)


.github/workflows/rust.yml
```diff
@@ -44,3 +44,12 @@ jobs:
 
       - name: Test CLI
         run: make test-cli
+
+      - name: Build resolc
+        run: make install-bin
+
```
xermicus, line 50: This should not be required? Since the test-* targets do it anyways

smiasojed: Right, fixed

PR #42, opened 2024-09-10
[solidity] Fix wrong check on recursive-process-input CLI option.

wpt967:
Fix the check for the validity of `--recursive-process-input` to make sure that it is only selected when `--recursive-process` is also selected.


xermicus:
Sorry. My brain must have short circuited when I fixed that clippy lint

PR #55, opened 2024-09-27
[workflows] Add workflow for building revive in a debian container.

wpt967:
Makefile: Add target 'install-revive' to build revive with the installation path specified by variable REVIVE_INSTALL_DIR.

Add utils directory with scripts for building revive in a container.

Add utils/build-revive.sh taking option argument '-o <install-dir>' to build revive with the specified install directory.

Add utils/revive-builder-debian.dockerfile to make a docker container for building revive in a Debian environment.

### DIFF
--- a/.github/workflows/build-revive-debian.yml
+++ b/.github/workflows/build-revive-debian.yml
@@ -1,0 +1,38 @@
+name: Build revive-debian
+on:
+  workflow_dispatch:
+
+env:
+  REVIVE_DEBIAN_PACKAGE: revive-debian-x86
+  DEBIAN_CONTAINER: revive-builder-debian-x86
+  DEBIAN_CONTAINER_BUILDER: build-debian-builder.sh
+  DEBIAN_CONTAINER_RUNNER: run-debian-builder.sh
+  REVIVE_DEBIAN_INSTALL: ${{ github.workspace }}/target/release
+  REVIVE_DEBIAN_BINARY: resolc
+  RUST_VERSION: "1.80"
+
+jobs:
+  build-revive-debian-x86:
+    name: debian-container-x86
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: build-container
+        run: |
+          (cd utils && ./${{ env.DEBIAN_CONTAINER_BUILDER}} --build-arg RUST_VERSION=${{ env.RUST_VERSION}} . )
+
+      - name: build-revive-debian
+        run: |
+          rustup show
+          cargo --version
+          rustup +nightly show
+          cargo +nightly --version
+          bash --version
+          utils/${{ env.DEBIAN_CONTAINER_RUNNER }} utils/build-revive.sh -o ${{ env.REVIVE_DEBIAN_INSTALL}}
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ env.REVIVE_DEBIAN_PACKAGE }}
+          path: ${{ env.REVIVE_DEBIAN_INSTALL }}/${{ env.REVIVE_DEBIAN_BINARY }}
+          retention-days: 1
--- a/utils/build-revive.sh
+++ b/utils/build-revive.sh
@@ -1,0 +1,20 @@
+#! /usr/bin/env bash
+
+set -euo pipefail
+
+REVIVE_INSTALL_DIR=$(pwd)/target/release
+while getopts "o:" option ; do
+    case $option in
+    o) # Output directory
+        REVIVE_INSTALL_DIR=$OPTARG
+        ;;
+    \?) echo "Error: Invalid option"
+        exit 1;;
+    esac
+done
+echo "Installing to ${REVIVE_INSTALL_DIR}"
+
+$(pwd)/build-llvm.sh
+export PATH=$(pwd)/llvm18.0/bin:$PATH
+
+make install-revive REVIVE_INSTALL_DIR=${REVIVE_INSTALL_DIR}
--- a/utils/revive-builder-debian.dockerfile
+++ b/utils/revive-builder-debian.dockerfile
@@ -1,0 +1,14 @@
+# syntax=docker/dockerfile:1
+# Dockerfile for building revive in a Debian container.
+FROM debian:12
+RUN <<EOF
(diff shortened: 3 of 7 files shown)


.github/workflows/build-revive-debian.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  build-revive-debian-x86:
+    name: debian-container-x86
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: build-container
+        run: |
+           (cd utils && ./${{ env.DEBIAN_CONTAINER_BUILDER}} . )
```
smiasojed: why did you decide to use your own docker. With this [ci-unified] (https://github.com/paritytech/scripts/blob/master/dockerfiles/ci-unified/README.md)
you have rust already installed and you can control the versions using docker tags.
The only issue which I see is that they are using debian 11 and remix backend is based on debian 12.

.github/workflows/build-revive-debian.yml
```diff
@@ -0,0 +1,32 @@
...
+    name: debian-container-x86
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: build-container
+        run: |
+           (cd utils && ./${{ env.DEBIAN_CONTAINER_BUILDER}} --build-arg RUST_VERSION=${{ env.RUST_VERSION}} . )
+
+      - name: build-revive-debian
```
smiasojed, line 25: I think it would be good to print the versions of the tools being used. Please take a look [here](https://github.com/use-ink/ink/blob/master/.github/rust-info/action.yml)

.github/workflows/build-revive-debian.yml
```diff
@@ -0,0 +1,32 @@
...
+  workflow_dispatch:
+
+env:
+  REVIVE_DEBIAN_PACKAGE: revive-debian-x86
+  DEBIAN_CONTAINER: revive-builder-debian-x86
+  DEBIAN_CONTAINER_BUILDER: build-debian-builder.sh
+  DEBIAN_CONTAINER_RUNNER: run-debian-builder.sh
+  REVIVE_DEBIAN_INSTALL: ${{ github.workspace }}/target/release
+  REVIVE_DEBIAN_BINARY: resolc
+  RUST_VERSION: stable
```
xermicus, line 12: The problem here is `stable` will change over time, but we want to have manual control over it. We can stick to 1.80 (slightly older version should do fine too).

utils/revive-builder-debian.dockerfile
```diff
@@ -0,0 +1,14 @@
+# syntax=docker/dockerfile:1
+# Dockerfile for building revive in a Debian container.
+FROM debian:12
+RUN <<EOF
+apt-get update
+apt-get install -q -y build-essential cmake make ninja-build python3 \
+    libmpfr-dev libgmp-dev libmpc-dev ncurses-dev \
+    git curl
+EOF
+ARG RUST_VERSION=stable
```
xermicus, line 10: Same here, I think sticking to specific version should be better.

What is missing is a rust version in the top level `Cargo.toml`:
```
rust-version = "1.80.0"
```

Could you add this here and to the top level Cargo.toml?

utils/build-revive.sh
```diff
@@ -0,0 +1,20 @@
...
+set -euo pipefail
+
+REVIVE_INSTALL_DIR=$(pwd)/target/release
+while getopts "o:" option ; do
+    case $option in
+    o) # Output directory
+        REVIVE_INSTALL_DIR=$OPTARG
+        ;;
+    \?) echo "Error: Invalid option"
+        exit;;
```
xermicus, line 12: Should probably return non-zero value here?

```suggestion
        exit 1;;
```

xermicus (approved):
LGTM

.github/workflows/build-revive-debian.yml
xermicus: This indicates that the remix backend does also not use `ci-unifed`? But it's a good point, the debian version in the builder should probably match the version where the binary is run.

Note that if we settle on ci-unifed the supported rust version is 1.77 which is quite old. It'll probably able to build revive but I didn't try that.

.github/workflows/build-revive-debian.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  build-revive-debian-x86:
+    name: debian-container-x86
+    runs-on: ubuntu-latest
```
smiasojed, line 17: If the run takes more than half an hour, maybe we should use beefier runners. I remember we had `ubuntu-latest-8-cores` and `ubuntu-latest-16-cores` runners.

xermicus: Yes we should use the more beefy ones. But building `resolc` itself takes seconds to minutes even on potato machines. I guess the main problem is LLVM and this is solved by having a dedicated release workflow for LLVM. So I think we should try both: Beefy runners + more economic release workflows.

.github/workflows/build-revive-debian.yml
wpt967: What is the issue with installing rust in a Debian container?

.github/workflows/build-revive-debian.yml
wpt967: Done

.github/workflows/build-revive-debian.yml
wpt967: Changed to 1.80

utils/revive-builder-debian.dockerfile
wpt967: The value in the Dockerfile is a default value, the value used is set in in the workflow file.

Changed the Cargo.toml file to specify the rust-version is 1.80

utils/build-revive.sh
wpt967: Done

.github/workflows/build-revive-debian.yml
smiasojed: I don't see any issue; I just prefer to avoid maintaining additional tools/images if I have something ready to use.

smiasojed (approved):

PR #70, opened 2024-10-09
implement immutable variables codegen

xermicus:
This PR implements immutable variables in the codegen backend.

Notes on the implementation (this also differences from the ZKSync implementation):

- Do not explicitly store the index. The stored index is used to calculate the index itself. It follows that it can always be implied.
- We only store the final immutable data size (and the required memory buffer length) after we finished translating to LLVM IR. This way the observed immutable data size is not depended of the position of translated returns and immutable accesses.
- Do not return immutable data from the constructor; instead use the syscall.
- Lazily load the whole immutable data upon the first access. Once the data is loaded, just read from memory. This works by only syscalling into `get_immutable_data` if the immutable data size global variable is not 0 and writing back 0 into this global variable after the first (and only) call to `get_immutable_data`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -758,7 +758,7 @@ checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 [[package]]
 name = "asset-test-utils"
 version = "7.0.0"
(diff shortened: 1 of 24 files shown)


smiasojed:
Could we increment the revive version? Most likely, we will do the next dev release soon with this change.

xermicus:
We are not publishing anything to crates.io yet (I think we can't anyways until we get rid of the polkadot-sdk git repo)

wpt967 (approved):
OK

PR #71, opened 2024-10-09
[llvm-context,solidity] Add a command line option to output LLVM-IR

wpt967:
Add option '--output-llvm-ir' to output LLVM-IR for each contract. This is only available in debug builds. It cannot be used with '--asm' or '--bin'.

### DIFF
--- a/crates/llvm-context/src/polkavm/context/mod.rs
+++ b/crates/llvm-context/src/polkavm/context/mod.rs
@@ -85,6 +85,10 @@ where
     dependency_manager: Option<D>,
     /// Whether to append the metadata hash at the end of bytecode.
     include_metadata_hash: bool,
+
+    /// Whether to emit LLVM-IR only.
+    output_llvm_ir: bool,
+
     /// The debug info of the current module.
     // debug_info: DebugInfo<'ctx>,
     /// The debug configuration telling whether to dump the needed IRs.
@@ -196,6 +200,7 @@ where
         optimizer: Optimizer,
         dependency_manager: Option<D>,
         include_metadata_hash: bool,
+        output_llvm_ir: bool,
         debug_config: Option<DebugConfig>,
     ) -> Self {
         Self::link_stdlib_module(llvm, &module);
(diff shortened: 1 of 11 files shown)


xermicus (commented):
There is already a `--debug-output-dir` argument, which leaves (per contract) both the LLVM IR before and after optimizations in the specified debug output directory

wpt967:
> There is already a `--debug-output-dir` argument, which leaves (per contract) both the LLVM IR before and after optimizations in the specified debug output directory. Not sure if we need this extra option?

That is a debugging option that produces the LLVM-IR as a side effect. It is also not clear what status the two files have. The new option makes LLVM-IR an explicit output format of the compiler and it is clearer that the output is the final code.

xermicus:
> That is a debugging option that produces the LLVM-IR as a side effect. 

I'm not sure if I understand. The help of `--debug-output-dir` says: `Dump all IRs to files in the specified directory. Only for testing and debugging`. To me this reads like dumping the LLVM is not just a mere side effect but the intended behavior? Or dumping the LLVM artifacts is a side effect of what exactly?

> It is also not clear what status the two files have. The new option makes LLVM-IR an explicit output format of the compiler and it is clearer that the output is the final code.

I don't understand what you mean with the "status" of the file? AFAICT, the only thing the option added in this PR does is just partially the same thing as `--debug-output-dir` does (dumping the LLVM IR _after_ optimization into the wrong directory). FWIW, arguably `--debug-output-dir` dumping before and after optimizations and name them accordingly is much "cleearer".

Also, this new flag would be (rightfully so) only available in debug mode. Which means it is `Only for testing and debugging`. Just like the existing flag. I can't understand whats wrong with the existing flag and I do not understand the use case for this. To get the debug artifacts in the same output dir, just specify the same dir as the output dir.

xermicus:
And it would also require adding more distinct flags akin to this to dump all stages (unoptimized, YUL, etc.), otherwise we just have multiple different ways to do the same thing. Which I'd like to avoid.

wpt967:
This seemed like a useful option but it can be done using the existing options. I'll close this PR.

xermicus:
Glad that we can settle on using the existing option.

FWIW an enhancement to the existing option would be to allow filtering which artifacts we want. But let's not make such cosmetics a priority for now :) Thanks!

PR #74, opened 2024-10-10
[llvm-context, solidity] Add CLI option '-g' to generate debug inform…

ghost:
…ation.

Add command line option '-g' to generate source level debug information in the output code. This only works with the LLVM-IR code generator.

Add flag 'emit_debug_info' to the llvm-context optimizer/code-gen settings structure, to record the setting of the '-g' CLI option.

Generate source level debug information for the functions defined when YUL is lowered to LLVM-IR. This includes the deploy_code and runtime functions generated by the compiler.

Generate debug-location information for other constructs that may appear in a contract.

### DIFF
--- a/crates/solidity/src/tests/cli-tests/tests/common.test.ts
+++ b/crates/solidity/src/tests/cli-tests/tests/common.test.ts
@@ -77,56 +77,68 @@ describe("Default run a command from the help", () => {
     });
 });
 
-describe("Run resolc with source debug information enabled", () => {
+describe("Run resolc with source debug information", () => {
+    const commands = [
+        `resolc -g ${paths.pathToBasicSolContract}  --bin --asm --output-dir "${paths.pathToOutputDir}"`,
+        `resolc --disable-solc-optimizer -g ${paths.pathToBasicSolContract}  --bin --asm --output-dir "${paths.pathToOutputDir}"`
+    ]; // potential issue on resolc with full path on Windows cmd`;
 
-    const command = `resolc -g ${paths.pathToBasicSolContract}  --bin --asm --output-dir "${paths.pathToOutputDir}"`; // potential issue on resolc with full path on Windows cmd`;
-    const result = executeCommand(command);
+    for (var idx in commands) {
+        const command = commands[idx];
+        const result = executeCommand(command);
 
-    it("Compiler run successful", () => {
-        expect(result.output).toMatch(/(Compiler run successful.)/i);
-    });
-    it("Exit code = 0", () => {
-        expect(result.exitCode).toBe(0);
-    });
-    it("Output dir is created", () => {
-        expect(isFolderExist(paths.pathToOutputDir)).toBe(true);
-    });
-    xit("Output files are created", () => { // a bug on windows
-        expect(isFileExist(paths.pathToOutputDir, paths.contractSolFilename, paths.binExtension)).toBe(true);
-        expect(isFileExist(paths.pathToOutputDir, paths.contractSolFilename, paths.asmExtension)).toBe(true);
-    });
-    it("the output files are not empty", () => {
-        expect(isFileEmpty(paths.pathToSolBinOutputFile)).toBe(false);
-        expect(isFileEmpty(paths.pathToSolAsmOutputFile)).toBe(false);
-    });
-    it("No 'Error'/'Fail' in the output", () => {
-        expect(result.output).not.toMatch(/([Ee]rror|[Ff]ail)/i);
-    });
+        it("Compiler run successful", () => {
+            expect(result.output).toMatch(/(Compiler run successful.)/i);
+        });
+        it("Exit code = 0", () => {
+            expect(result.exitCode).toBe(0);
+        });
+        it("Output dir is created", () => {
+            expect(isFolderExist(paths.pathToOutputDir)).toBe(true);
+        });
+        it("Output files are created", () => { // a bug on windows
+            expect(isFileExist(paths.pathToOutputDir, paths.contractSolFilename, paths.binExtension)).toBe(true);
+            expect(isFileExist(paths.pathToOutputDir, paths.contractSolFilename, paths.asmExtension)).toBe(true);
+        });
+        it("the output files are not empty", () => {
+            expect(isFileEmpty(paths.pathToSolBinOutputFile)).toBe(false);
+            expect(isFileEmpty(paths.pathToSolAsmOutputFile)).toBe(false);
+        });
+        it("No 'Error'/'Fail' in the output", () => {
+            expect(result.output).not.toMatch(/([Ee]rror|[Ff]ail)/i);
+        });
+    }
 });
 
-describe("Run resolc with source debug information and no solc optimization", () => {
+describe("Run resolc with source debug information, check LLVM debug-info", () => {
+    const commands = [
+        `resolc -g ${paths.pathToBasicSolContract} --debug-output-dir="${paths.pathToOutputDir}"`,
+        `resolc -g --disable-solc-optimizer ${paths.pathToBasicSolContract} --debug-output-dir="${paths.pathToOutputDir}"`
+    ]; // potential issue on resolc with full path on Windows cmd`;
 
-    const command = `resolc --disable-solc-optimizer -g ${paths.pathToBasicSolContract}  --bin --asm --output-dir "${paths.pathToOutputDir}"`; // potential issue on resolc with full path on Windows cmd`;
-    const result = executeCommand(command);
+    for (var idx in commands) {
+        const command = commands[idx];
+        const result = executeCommand(command);
 
-    it("Compiler run successful", () => {
-        expect(result.output).toMatch(/(Compiler run successful.)/i);
-    });
-    it("Exit code = 0", () => {
-        expect(result.exitCode).toBe(0);
-    });
-    it("Output dir is created", () => {
-        expect(isFolderExist(paths.pathToOutputDir)).toBe(true);
-    });
-    xit("Output files are created", () => { // a bug on windows
-        expect(isFileExist(paths.pathToOutputDir, paths.contractSolFilename, paths.binExtension)).toBe(true);
-        expect(isFileExist(paths.pathToOutputDir, paths.contractSolFilename, paths.asmExtension)).toBe(true);
-    });
-    it("the output files are not empty", () => {
-        expect(isFileEmpty(paths.pathToSolBinOutputFile)).toBe(false);
-        expect(isFileEmpty(paths.pathToSolAsmOutputFile)).toBe(false);
-    });
-    it("No 'Error'/'Fail' in the output", () => {
-        expect(result.output).not.toMatch(/([Ee]rror|[Ff]ail)/i);
-    });
+        it("Compiler run successful", () => {
+            expect(result.output).toMatch(/(Compiler run successful.)/i);
+        });
+        it("Exit code = 0", () => {
+            expect(result.exitCode).toBe(0);
+        });
+        it("Output dir is created", () => {
+            expect(isFolderExist(paths.pathToOutputDir)).toBe(true);
+        });
+        it("Output files are created", () => { // a bug on windows
+            expect(isFileExist(paths.pathToOutputDir, paths.contractOptimizedLLVMFilename, paths.llvmExtension)).toBe(true);
+            expect(isFileExist(paths.pathToOutputDir, paths.contractUnoptimizedLLVMFilename, paths.llvmExtension)).toBe(true);
+        });
+        it("the output files are not empty", () => {
+            expect(isFileEmpty(paths.pathToSolBinOutputFile)).toBe(false);
+            expect(isFileEmpty(paths.pathToSolAsmOutputFile)).toBe(false);
+        });
+        it("No 'Error'/'Fail' in the output", () => {
+            expect(result.output).not.toMatch(/([Ee]rror|[Ff]ail)/i);
+        });
+    }
 });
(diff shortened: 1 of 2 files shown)


wpt967:
Add debug-info to the immutables-load function.

xermicus (changes requested):
As a rule of thumb, repeated code-gen logic should be implemented as methods to the context and this applies to debug info as well. I suggest basically any `if let Some(dinfo) = context.debug_info()` be a method `context.debug_info_something` instead. They can all return early if debug info isn't requested. This will 1) solve the issue of the heavy code duplication, 2) greatly ease the reading flow of the implementation as all the debug info related stuff can be skipped by skipping just over a single line of code and 3) make it consistent with the regular code-gen related code.

Can you please also enable debug info in the integration test suite by default to get test coverage?

Thanks

crates/solidity/src/project/contract/mod.rs
```diff
@@ -154,6 +167,19 @@ impl Contract {
             )
         })?;
 
+        if let Some(dinfo) = context.debug_info() {
+            dinfo.finalize_module()
+        }
+
+        #[cfg(debug_assertions)]
+        if let Err(err) = context.verify() {
+            eprintln!("Invalid LLVM module generated:");
```
xermicus, line 176: FIY error context should be added using the anyhow context. In stderr it might also be swallowed in recursive calls.

wpt967: Fixed.

crates/solidity/src/project/contract/mod.rs
```diff
@@ -154,6 +167,19 @@ impl Contract {
             )
         })?;
 
+        if let Some(dinfo) = context.debug_info() {
+            dinfo.finalize_module()
+        }
+
+        #[cfg(debug_assertions)]
```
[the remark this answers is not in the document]
wpt967: Removed.

crates/llvm-context/src/polkavm/context/function/runtime/runtime_code.rs
```diff
@@ -59,7 +61,62 @@ where
 
         context.set_basic_block(context.current_function().borrow().entry_block());
         context.set_code_type(CodeType::Runtime);
+        if let Some(dinfo) = context.debug_info() {
```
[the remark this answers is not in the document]
wpt967: Add functions to build and setup function and location debug-info.

crates/llvm-context/src/polkavm/context/function/runtime/immutable_data_load.rs
```diff
@@ -37,6 +39,53 @@ where
         context.set_current_function(runtime::FUNCTION_LOAD_IMMUTABLE_DATA)?;
         context.set_basic_block(context.current_function().borrow().entry_block());
 
+        if let Some(dinfo) = context.debug_info() {
+            let di_builder = dinfo.builder();
+            let func_name: &str = runtime::FUNCTION_RUNTIME_CODE;
```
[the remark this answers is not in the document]
wpt967: Fixed.

crates/llvm-context/src/polkavm/context/function/runtime/deploy_code.rs
```diff
@@ -60,7 +62,57 @@ where
...
+            let func_name: &str = runtime::FUNCTION_DEPLOY_CODE;
+            let di_file = dinfo.compilation_unit().get_file();
+            let di_scope = dinfo.top_scope().expect("expected a debug-info scope");
+
+            let di_flags = inkwell::debug_info::DIFlagsConstants::PUBLIC;
+            let ret_type = dinfo.create_word_type(Some(di_flags))?.as_type();
+            let subroutine_type =
+                di_builder.create_subroutine_type(di_file, Some(ret_type), &[], di_flags);
+            let linkage = dinfo.namespace_as_identifier(Some(func_name));
+            let di_func_scope = di_builder.create_function(
```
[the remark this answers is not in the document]
wpt967: Fixed in the updated code.

crates/llvm-context/src/polkavm/context/function/runtime/runtime_code.rs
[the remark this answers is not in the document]
xermicus: Nice, I glanced over it, looks much more concise, thanks!

wpt967:
Reworked 'set_debug_location' to make it easier to set debug locations.

crates/llvm-context/src/polkavm/context/function/runtime/immutable_data_load.rs
```diff
@@ -37,6 +39,15 @@ where
...
 
+        if context.debug_info().is_some() {
+            context.builder().unset_current_debug_location();
+            let func_scope = context
+                .set_current_function_debug_info(runtime::FUNCTION_LOAD_IMMUTABLE_DATA, 0)?
+                .as_debug_info_scope();
+            context.debug_info().unwrap().push_scope(func_scope);
+            context.set_debug_location(0, 0, Some(func_scope))?;
+        }
+
```
xermicus, line 50: This pairs with `set_current_function`? Setting the current function implies adjusting the debug info location. So it should be done in `fn set_current_function`.

This also reduces the potential to forget about it. For example, in the EVMLA codegen it is missing.

crates/llvm-context/src/polkavm/context/function/runtime/immutable_data_load.rs
```diff
@@ -111,6 +122,10 @@ where
         context.set_basic_block(return_block);
         context.build_return(None);
 
+        if let Some(dinfo) = context.debug_info() {
```
xermicus, line 125: This is inherently tied to `fn set_current_function`.

Whenever we do this for a function we have an inconsistency. Because we set the current function, push a di scope, pop a di scope, but do not unset the current function.

Which implies that at any point, we assume to have a function set (because there is no way to unset it).

Which implies that this can and should be done in `fn set_current_function` too, before pushing the new scope.

(This and the other comment should apply for functions in general)

crates/llvm-context/src/polkavm/context/function/runtime/entry.rs
```diff
@@ -221,14 +225,32 @@ where
...
+                .as_debug_info_scope();
+            context.debug_info().unwrap().push_scope(func_scope);
+            context.set_debug_location(0, 0, Some(func_scope))?;
+        }
+
         Self::initialize_globals(context)?;
+        context.set_debug_location(0, 0, None)?;
+
         Self::load_calldata(context)?;
+        context.set_debug_location(0, 0, None)?;
```
xermicus, line 241: This should not be needed? AFAICT we never left the basic block so far

crates/llvm-context/src/polkavm/context/function/runtime/entry.rs
```diff
@@ -221,14 +225,32 @@ where
...
+
         Self::leave_entry(context)?;
+        context.set_debug_location(0, 0, None)?;
```
xermicus, line 244: Same here, why are those needed?

crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -480,6 +498,84 @@ where
...
+    }
+
+    pub fn set_debug_location(
+        &self,
+        line: usize,
+        column: usize,
+        scope: Option<DIScope<'ctx>>,
+    ) -> anyhow::Result<()> {
+        if let Some(dinfo) = self.debug_info() {
+            let line_num: u32 = std::cmp::min(line, u32::MAX as usize) as u32;
```
xermicus, line 561: We have a lot of those. A line number greater than u32::MAX should never occur. I doubt solc does emit this. So this should be try_into?

However, the parser must handle this, i.e. error out if reported line and column numbers make no sense. During codegen we expect a syntactically correct program. And then this is no longer necessary.

xermicus (commented):
> Reworked 'set_debug_location' to make it easier to set debug locations.

Thanks! I really like it. Half of the lines added are gone :) However there are still some inconsistencies I'd like to iron out.

Also the EVMLA is rather untouched yet, I think it requires some changes to emit correct debug info too.

crates/solidity/src/yul/parser/statement/function_definition.rs
```diff
@@ -265,13 +267,61 @@ where
         context: &mut revive_llvm_context::PolkaVMContext<D>,
     ) -> anyhow::Result<()> {
         context.set_current_function(self.identifier.as_str())?;
-        let r#return = context.current_function().borrow().r#return();
```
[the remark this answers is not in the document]
wpt967: This has been moved to where it is first used.

crates/solidity/src/project/contract/mod.rs
wpt967: Fixed

crates/llvm-context/src/polkavm/context/function/runtime/immutable_data_load.rs
wpt967: The EVMLA will be done separately, this PR is concerned with the LLVM part of the pipeline.

xermicus: Well, EVMLA code is lowered to LLVM IR by mostly the same backend code. But EVMLA is of lower priority than YUL. As long as the EVMLA pipeline does not break as a result of this PR, I'm ok with doing it later.

crates/llvm-context/src/polkavm/context/function/runtime/immutable_data_load.rs
xermicus: I mean if this doesn't work. I might be missing something. But if we mismatch the code we emit with the function scope we emit, it smells like something should be broken. WDYT?

wpt967: Moved function debug-info generation into `set_current_function`.

xermicus: Ok that might not work at all. We might have to stick with the implication that we are done at the end here.

xermicus: > Ok that might not work at all. We might have to stick with the implication that we are done at the end here.

(This was addressing my own comment I wrote earlier above)

> Moved function debug-info generation into set_current_function.

Sounds good. Assuming this has to be done anyways when setting the function, seems like the right place :) I guess popping the context probably needs to inherently stay separate. For now I'm fine with sprinkling `context.pop_debug_scope();` everywhere we expect to not come back emitting more code.

Cargo.toml
```diff
@@ -51,10 +51,10 @@ path-slash = "0.2"
 rayon = "1.8"
 clap = { version = "4", default-features = false, features = ["derive"] }
 rand = "0.8"
-polkavm-common = "0.10"
-polkavm-linker = "0.10"
-polkavm-disassembler = "0.10"
-polkavm = "0.10"
+polkavm-common = "0.12"
```
xermicus, line 54: I don't get it - this is already in `paritytech:main`? :D

xermicus: maybe this goes away after a rebase against main?

wpt967: This is part of a commit on the main branch that was accidentally included.

crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -471,15 +485,133 @@ where
...
+        Ok(())
+    }
+
+    /// Builds a debug-info scope for a function.
+    pub fn build_function_debug_info(
+        &self,
+        name: &str,
+        line_no: u32,
+    ) -> anyhow::Result<inkwell::debug_info::DISubprogram<'ctx>> {
+        let dinfo = self.debug_info().expect("expected debug-info builder");
```
xermicus, line 511: Why is this not a `let else` with an early return?

crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -471,15 +485,133 @@ where
...
+                None,
+            );
+            self.builder().set_current_debug_location(di_loc);
+        }
         Ok(())
     }
 
+    /// Pushes a debug-info scope to the stack.
+    pub fn push_debug_scope(&self, scope: DIScope<'ctx>) {
+        if self.debug_info().is_some() {
```
xermicus, line 587: Should prefer `if let` instead of `is_some()` followed by unwraps when we operate on optional values only if they are `Some` (noticed this in other places too).

crates/llvm-context/src/polkavm/context/function/runtime/immutable_data_load.rs
```diff
@@ -34,9 +36,17 @@ where
...
         context.set_basic_block(context.current_function().borrow().entry_block());
 
+        if context.debug_info().is_some() {
+            context.builder().unset_current_debug_location();
+            let func_scope = context
+                .set_current_function_debug_info(runtime::FUNCTION_LOAD_IMMUTABLE_DATA, 0)?
+                .as_debug_info_scope();
+            context.push_debug_scope(func_scope);
+        }
+
```
xermicus, line 49: This code is duplicated 3x. Since it is needed every time we lower functions it should be a helper on context instead.

Cargo.toml
xermicus: Ok, can you please rebase so we get a clean diff?

crates/solidity/src/tests/cli-tests/tests/common.test.ts
```diff
@@ -76,3 +76,57 @@ describe("Default run a command from the help", () => {
         expect(result.output).not.toMatch(/([Ee]rror|[Ww]arning|[Ff]ail)/i);
     });
 });
+
+describe("Run resolc with source debug information enabled", () => {
```
xermicus, line 80: Can we verify in this test that some of the expected debug info is in the actual final code blob? Doesn't need to be fancy, a string match or something we expect only to appear with debug info would be sufficient.

xermicus (commented):
Debug info generation is not enabled by default in our integration test right? All test pass when I switch the default behavior to always generate debug info, so we should be good to enable it.

PR #76, opened 2024-10-10
disable sbrk and emulate EVM linear memory internally

xermicus:
This PR gets rid of calling the PVM sbrk functionality because it was disabled in the pallet. Instead we emulate the EVM linear memory heap internally with a static memory buffer.

I also changed the memory tests to run differentially; I found no reason why we our linear memory implementation needs to differ in intentional ways.


wpt967 (approved):
LGTM

Issue #63, opened 2024-10-04
Support the `extcodehash` opcode


xermicus:
Needs a supporting API method in the pallet.

xermicus:
Assigned to @ermalkaleci

xermicus:
To be more precise, check the comments in #77:

@ermalkaleci The pallet should be changed to write the zero hash if the account exists but has no code, and all zeroes if the account was not found. The status code can then either be removed (since the callee can check it by inspecting the output buffer after the call value) or changed to reflect what exactly was written accordingly.

ermalkaleci:
@xermicus this is resolved

xermicus:
Thank you!

PR #77, opened 2024-10-11
Implement extcodehash

ermalkaleci:
- Add manual tests for `extcodehash` because we can't hardcode the hash
- Update specs instantiate result to include code hash

TODO:
- [x] Update pallet-revive to return hardcoded hash for EOA


crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -666,6 +666,18 @@ where
             .into())
     }
 
+    /// Load bytes32 at given pointer
+    pub fn build_load_bytes(
```
xermicus, line 670: This helper is not needed. It boils down to just `self.build_byte_swap(self.build_load(pointer, "bytes_pointer")?)` since bytes32 size equals the word size?

crates/llvm-context/src/polkavm/evm/ext_code.rs
```diff
@@ -45,11 +45,30 @@ where
...
+    let address_type = context.integer_type(revive_common::BIT_LENGTH_ETH_ADDRESS);
+    let address_pointer = context.build_alloca_at_entry(address_type, "address_pointer");
+    let address_truncated =
+        context
+            .builder()
+            .build_int_truncate(address, address_type, "address_truncated")?;
+    let address_swapped = context.build_byte_swap(address_truncated.into())?;
+    context.build_store(address_pointer, address_swapped)?;
+
+    let extcodehash_pointer = context.build_alloca(context.word_type(), "extcodehash_pointer");
```
xermicus, line 63: Can you please use `build_alloca_at_entry` here instead? The following load makes it safe.

crates/llvm-context/src/polkavm/const/runtime_api.rs
```diff
@@ -27,6 +27,8 @@ pub mod imports {
 
     pub static CODE_SIZE: &str = "code_size";
 
+    pub static CODE_HASH: &str = "code_hash";
```
xermicus, line 30: Should be added to the list below.

crates/integration/contracts/ExtCode.sol
```diff
@@ -2,6 +2,64 @@
...
+                "dest": {
+                    "Instantiated": 0
+                },
+                "data": "cbb20918"
+            }
+        },
+        {
+            "VerifyCall": {
+                "success": true,
+                "output": "10b3daf252aa843eae2b3d850da130d6a41ce1c23f33081c7fd8e48a0844add8"
```
xermicus, line 28: I fear this makes the test very brittle. Those JSON specs are intended for differential testing against EVM. But those code hashes are different than the corresponding EVM code bytecode hashes. And they change all the time, we don't want to manually update them each time the code hash of the fixture changes (I also want to avoid maintaining compiled fixtures).

I think the easiest way to write those tests without them breaking all the time is is to just write them iteratively. Basically just add the contract [like this](https://github.com/paritytech/revive/blob/main/crates/integration/src/cases.rs#L209C1-L216C66). Then you can easily implement the test manually somewhere [here](https://github.com/paritytech/revive/blob/main/crates/integration/src/tests.rs). The alternative is to make to extend the spec runner format and runner so that we can express in the spec JSON that the we want to compare against the bytecode hash of a contract. Which might be a lot more work than just writing this test iteratively :) And we have planned changes to the test spec format anyways. So just doing it iteratively would be fine for me.

crates/llvm-context/src/polkavm/evm/ext_code.rs
```diff
@@ -45,11 +45,30 @@ where
...
+
+    context.build_runtime_call(
```
xermicus, line 65: This does not fully resemble the [EVM semantics](hash: hash of the chosen account's [code](https://www.evm.codes/about), the empty hash (0xc5d24601...) if the account has no [code](https://www.evm.codes/about), or 0 if the account does not exist or has been destroyed.). Here is [the syscall](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/revive/src/wasm/runtime.rs#L1409) for reference.

I think with the current implementation in `pallet-revive`, we are basically forced to emit some branches and make two syscalls for this. Which is not desired, everything the runtime can do for us easily it should do (it is faster and spares contract code size). So this needs a PR in the pallet first to let it distinguish between contract-accounts, no-contract accounts and unknown accounts and write back the correct value automatically. Then it can be properly implemented and tested.

Sorry in the corresponding issue #63 I should have wrote that out more clearly.

xermicus (changes requested):
Thanks for your pull request!

xermicus:
Closes #63. @ermalkaleci if you feel like takling this on the runtime side too, can I assign #63 to you?

crates/integration/contracts/ExtCode.sol
ermalkaleci: Yes I plan to do that

ermalkaleci:
Absolutely, I would love to contribute. Hope this is good start :)

xermicus:
Nice, thank you! Yeah we can take it from here. There's nothing complex to it, it is a good first issue.

ermalkaleci:
@xermicus can you give [this](https://github.com/paritytech/polkadot-sdk/pull/6088) a review?

crates/integration/src/tests.rs
```diff
@@ -234,6 +234,100 @@ fn signed_remainder() {
...
+    .run();
+    let (addr, code_hash) = match results.first().cloned() {
+        Some(CallResult::Instantiate {
+            result, code_hash, ..
+        }) => (result.result.unwrap().addr, code_hash),
+        _ => panic!("instantiate contract failed"),
+    };
+
+    // code hash of itself
+    {
```
xermicus, line 255: What is the additional scope for?

xermicus (approved):
thanks

crates/integration/src/tests.rs
ermalkaleci: just a visual separation of test cases

ermalkaleci: Sure

xermicus: thanks

PR #79, opened 2024-10-11
Fix implementation for balance_of

ermalkaleci:
**Pull Request Summary**
- Fix `balance_of` implementation to use correct runtime api
- Add `build_address_argument_store` helper to be used for address arguments

### DIFF
--- a/crates/integration/src/tests.rs
+++ b/crates/integration/src/tests.rs
@@ -234,33 +234,6 @@ fn signed_remainder() {
     run_differential(actions);
 }
 
-#[test]
-fn balance_of() {
-    let mut actions = instantiate("contracts/Value.sol", "Value");
-
-    actions.push(Call {
-        origin: TestAddress::Alice,
-        dest: TestAddress::Instantiated(0),
-        value: 0,
-        gas_limit: None,
-        storage_deposit_limit: None,
-        data: Contract::value_balance_of(Address::from(CHARLIE.to_fixed_bytes())).calldata,
-    });
-
-    actions.push(VerifyCall(VerifyCallExpectation {
-        success: true,
-        // Charlie's balance without ExistentialDeposit
-        output: OptionalHex::from(U256::from(1_000_000_000 - 1_000).to_be_bytes_vec()),
-        gas_consumed: None,
-    }));
-
-    Specs {
-        actions,
-        ..Default::default()
-    }
-    .run();
-}
-
 /*
 // These test were implement for the mock-runtime and need to be ported yet.
 
@@ -288,32 +261,6 @@ fn create2_failure() {
         .call();
 
     assert_eq!(output.flags, ReturnFlags::Revert);
-}
-
-
-#[test]
-fn balance() {
-    let (_, output) = assert_success(&Contract::value_balance_of(Default::default()), false);
-
-    let expected = U256::ZERO;
-    let received = U256::from_be_slice(&output.data);
-    assert_eq!(expected, received);
-
-    let expected = U256::from(54589);
-    let (mut state, address) = State::new_deployed(Contract::value_balance_of(Default::default()));
-    state.accounts_mut().get_mut(&address).unwrap().value = expected;
-
-    let contract = Contract::value_balance_of(address);
-    let (_, output) = state
-        .transaction()
-        .with_default_account(&contract.pvm_runtime)
-        .calldata(contract.calldata)
-        .call();
-
-    assert_eq!(ReturnFlags::Success, output.flags);
-
-    let received = U256::from_be_slice(&output.data);
-    assert_eq!(expected, received)
 }
 
 #[test]
--- a/crates/integration/src/tests.rs
+++ b/crates/integration/src/tests.rs
@@ -234,6 +234,33 @@ fn signed_remainder() {
     run_differential(actions);
 }
 
+#[test]
+fn balance_of() {
+    let mut actions = instantiate("contracts/Value.sol", "Value");
+
+    actions.push(Call {
+        origin: TestAddress::Alice,
+        dest: TestAddress::Instantiated(0),
+        value: 0,
+        gas_limit: None,
+        storage_deposit_limit: None,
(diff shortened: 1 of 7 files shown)


crates/integration/src/tests.rs
```diff
@@ -234,6 +234,33 @@ fn signed_remainder() {
...
+
+    Specs {
+        actions,
+        ..Default::default()
+    }
+    .run();
+}
+
 /*
 // These test were implement for the mock-runtime and need to be ported yet.
```
xermicus, line 265: This is actually one of those yet-to-be ported :)

To easily test this against EVM, the `Value.sol` integration test just needs a slight change. Basically because we aren't able to express using the address of an instantiated account as calldata with the current spec data. Just move the assertions into Solidity:

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8;

/* runner.json
{
    "differential": true,
    "actions": [
        {
            "Upload": {
                "code": {
                    "Solidity": {
                        "contract": "ValueTester"
                    }
                }
            }
        },
        {
            "Instantiate": {
                "value": 1024,
                "code": {
                    "Solidity": {
                        "contract": "Value"
                    }
                }
            }
        },
        {
            "Call": {
                "dest": {
                    "Instantiated": 0
                },
                "value": 123,
                "data": "3fa4f245"
            }
        }
    ]
}
*/

contract ValueTester {
    constructor() payable {}

    function balance_self() public view returns (uint ret) {
        ret = address(this).balance;
    }
}

contract Value {
    constructor() payable {
        ValueTester tester = new ValueTester{value: msg.value}();

        // own account
        assert(address(this).balance == 0);

        // tester account
        assert(address(tester).balance == msg.value);
        assert(tester.balance_self() == msg.value);

        // non-existant account
        assert(address(0xdeadbeef).balance == 0);
    }

    function value() public payable returns (uint ret) {
        ret = msg.value;
    }
}
```

crates/integration/src/tests.rs
```diff
@@ -234,6 +234,33 @@ fn signed_remainder() {
     run_differential(actions);
 }
 
+#[test]
```
xermicus, line 237: We can test this deferentially against geth EVM (see comment below). Then this manual test isn't needed :)

Could you please also remove the old test implementation that is commented out further down? Thanks!

crates/llvm-context/src/polkavm/evm/ether_gas.rs
```diff
@@ -40,22 +40,26 @@ pub fn balance<'ctx, D>(
...
-    let address_pointer = context.build_alloca(context.word_type(), "address_pointer");
-    context.build_store(address_pointer, address)?;
+    let address_type = context.integer_type(revive_common::BIT_LENGTH_ETH_ADDRESS);
+    let address_pointer = context.build_alloca_at_entry(address_type, "address_pointer");
+    let address_truncated =
+        context
+            .builder()
+            .build_int_truncate(address, address_type, "address_truncated")?;
+    let address_swapped = context.build_byte_swap(address_truncated.into())?;
+    context.build_store(address_pointer, address_swapped)?;
```
xermicus, line 50: I see we are starting to need this in more and more places. It would be great to have a method `fn build_address_argument_store` on the context that does exactly this, returning the address `address_pointer`.

Now this isn't your fault; it is fine like that. I'll leave it up to you to decide if you want to implement the helper on the context. If you do all other occurrences should be changed to use the new helper instead.

crates/llvm-context/src/polkavm/evm/ether_gas.rs
```diff
@@ -40,22 +40,26 @@ pub fn balance<'ctx, D>(
...
 
+    let balance_pointer = context.build_alloca(context.word_type(), "balance_pointer");
```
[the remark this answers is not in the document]
xermicus: The reason is documented [LLVM performance guide](https://llvm.org/docs/Frontend/PerformanceTips.html#use-of-allocas). However we might as well just use a scratch buffer for those. Also need to think about zero-ing them out as this would give as the default zero value, which is assumed a lot in EVM, for free. A context helper taking care of this seems the most sensibly solution. But this is something for another PR :) So I'll leave it up to you.

xermicus (commented):
The implementation looks correct. Thanks for helping out!

crates/integration/src/tests.rs
xermicus: A nice side effect of doing the assertions in Solidity is that we don't even have to consider things like the ED. We just expect the contract to observe the exact same values than what it would see in the EVM reference implementation :)

ermalkaleci: That's a great idea. Will update

crates/llvm-context/src/polkavm/evm/ether_gas.rs
ermalkaleci: Updated

xermicus (approved):

crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -654,6 +654,21 @@ where
         Pointer::new(r#type, AddressSpace::Stack, pointer)
     }
 
+    /// Build address argument store point with bytes swapped and truncated.
```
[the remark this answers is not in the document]
ermalkaleci: I updated the comment while you were writing the comment :D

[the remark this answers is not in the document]
ermalkaleci: Will apply your suggestion

crates/llvm-context/src/polkavm/evm/ether_gas.rs
xermicus: Good stuff, thanks!

crates/llvm-context/src/polkavm/context/mod.rs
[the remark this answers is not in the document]
xermicus: lol

anyways, thanks for the PR! I'll merge given CI passes

PR #80, opened 2024-10-12
Implement delegate_call

### LINKED ISSUE #67: Support the `delegatecall` opcode
The API is already available in the pallet and should work fine. Once this is implemented we should test it with the openzeppelin proxy implementations.

ermalkaleci:
Closes #67

TODO:
- [x]  Update pallet-revive to take address instead of code hash


crates/integration/contracts/Delegate.sol
```diff
@@ -0,0 +1,80 @@
...
+}
+*/
+
+contract Logic {
+    // NOTE: storage layout must be the same as contract Tester
+    uint256 public num;
+    address public sender;
+    uint256 public value;
+
+    event DidSetVars();
```
xermicus, line 46: Testing this is irrelevant here? Keeping the fixtures as concise as possible eases debugging them :)

crates/integration/src/tests.rs
```diff
@@ -43,6 +43,7 @@ test_spec!(call, "Caller", "Call.sol");
 test_spec!(transfer, "Transfer", "Transfer.sol");
 test_spec!(return_data_oob, "ReturnDataOob", "ReturnDataOob.sol");
 test_spec!(immutables, "Immutables", "Immutables.sol");
+test_spec!(delegate, "Delegate", "Delegate.sol");
```
xermicus, line 46: This could use some better test coverage, i.e. calling into non-existent accounts, plain accounts, recursively into self (from runtime code and the constructor).

crates/llvm-context/src/polkavm/evm/call.rs
```diff
@@ -103,20 +103,79 @@ where
...
 where
     D: Dependency + Clone,
 {
-    todo!()
+    let address_pointer = context.build_address_argument_store(address)?;
+
+    let extcodehash_pointer =
+        context.build_alloca_at_entry(context.word_type(), "extcodehash_pointer");
+
+    context.build_runtime_call(
```
xermicus, line 124: I think with the current runtime this isn’t necessarily correct (might never be written to, so reading from it is potentially UB).

The reason `delegate_call` accepts a code hash instead of an address is probably so that people can upload code that they can delegate call into without having to instantiate first. But this isn’t expressible in Solidity (if it was it would be problematic because of the immutables - we assume to be instantiated first). Ideally for Solidity, the runtime API would take care of the code hash image lookup. Moving it into the runtime reduces contract code size and closes the door for related bugs in contracts. But it would need either a change or a new API (to preserve to use case of not requiring to instantiate new code hashes first). 

If we decide to stick with the current version of `delegate_call`, I’d like this PR to wait for #77 to avoid too much churn.

crates/llvm-context/src/polkavm/evm/call.rs
```diff
@@ -103,20 +103,79 @@ where
...
+    );
+
+    context.build_byte_swap(context.build_load(extcodehash_pointer, "extcodehash_value")?)?;
+
+    let input_offset = context.safe_truncate_int_to_xlen(input_offset)?;
+    let input_length = context.safe_truncate_int_to_xlen(input_length)?;
+    let output_offset = context.safe_truncate_int_to_xlen(output_offset)?;
+    let output_length = context.safe_truncate_int_to_xlen(output_length)?;
+
+    // TODO: What to supply here? Is there a weight to gas?
```
xermicus, line 139: We need some further investigation on this. Why is gas not a supported parameter in the pallet? Can it be supported? If so, that needs either an issue or PR for the pallet. Otherwise the gas argument should be removed from this function altogether and the reason documented.

crates/llvm-context/src/polkavm/evm/call.rs
```diff
@@ -103,20 +103,79 @@ where
...
+        .builder()
+        .build_int_truncate(_gas, context.integer_type(64), "gas")?;
+
+    let input_pointer = context.build_heap_gep(input_offset, input_length)?;
+    let output_pointer = context.build_heap_gep(output_offset, output_length)?;
+
+    let output_length_pointer = context.build_alloca_at_entry(context.xlen_type(), "output_length");
+    context.build_store(output_length_pointer, output_length)?;
+
+    let flags = context.xlen_type().const_int(0u64, false);
```
xermicus, line 150: Delegate calls can be reentrant.

crates/llvm-context/src/polkavm/evm/call.rs
```diff
@@ -103,20 +103,79 @@ where
...
+
+    context.build_runtime_call(
+        runtime_api::imports::CODE_HASH,
+        &[
+            address_pointer.to_int(context).into(),
+            extcodehash_pointer.to_int(context).into(),
+        ],
+    );
+
+    context.build_byte_swap(context.build_load(extcodehash_pointer, "extcodehash_value")?)?;
```
xermicus, line 132: Why is this needed?

crates/llvm-context/src/polkavm/evm/call.rs
```diff
@@ -103,20 +103,79 @@ where
 
 #[allow(clippy::too_many_arguments)]
 pub fn delegate_call<'ctx, D>(
-    _context: &mut Context<'ctx, D>,
+    context: &mut Context<'ctx, D>,
     _gas: inkwell::values::IntValue<'ctx>,
-    _address: inkwell::values::IntValue<'ctx>,
+    address: inkwell::values::IntValue<'ctx>,
     _value: Option<inkwell::values::IntValue<'ctx>>,
```
xermicus, line 109: Should be removed.

crates/llvm-context/src/polkavm/evm/call.rs
ermalkaleci: pallet-revive doesn't allow reentrant for `delegate_call` https://github.com/paritytech/polkadot-sdk/blob/d1c115b6197bf6c45d5640594f0432e6c2781a4f/substrate/frame/revive/src/wasm/runtime.rs#L1030

crates/llvm-context/src/polkavm/evm/call.rs
ermalkaleci: I would prefer we change pallet-revive to accept address rather than code hash. I was trying to work with what we have

crates/llvm-context/src/polkavm/evm/call.rs
xermicus: Ah okay. Would be good to know the reason for this and document it :) I mean ideally this would be changed to better match the EVM semantics.

crates/llvm-context/src/polkavm/evm/call.rs
xermicus: Yes the pallet doing the address lookup would definitively be preferred here.

FYI the pallet is under construction too, it isn't deployed anywhere. At that stage we are free to make breaking changes however we wish, as long as they are justified.

crates/llvm-context/src/polkavm/evm/call.rs
ermalkaleci: At first I thought pallet-revive was updated to fully support revive but now it seems more just a copy of pallet-contracts

crates/llvm-context/src/polkavm/evm/call.rs
xermicus: Maybe I should have stated that earlier. Both the compiler and the pallet are new, in the works and not yet deployed anywhere. But they are not entirely new code bases. We generally try to adapt the EVM behavior, as to introduce too many differences from what people expect on EVM :)

So the rule of thumb is, anything that differs should be investigated and changed. It doesn't mean that you have to do it all (even though we can use any helping hands) - if you feel like it is getting too tiring I can open corresponding polkadot-sdk issues and add them to the [tracking issue](https://github.com/paritytech/polkadot-sdk/issues/5308).

ermalkaleci: I thought pallet-revive will work with ink! and rust contracts and thought we need to somehow make revive work around that. If that's not that case then I can open PR on pallet-revive

xermicus: ink! will work fine on the revive-runtime since all changes can be reflected in a new version. We need the runtime to align with the incumbent platform and not vice-versa.

xermicus: > I can open PR on pallet-revive

Great, thanks! Just link this discussion there for context.  I'm un-opinionated how this should exactly look like - having distinct `delegate_call_by_hash` and `delegate_call_by_address` syscalls seems reasonable. Or just remove the ability to delegate call by hashes altogether if it is not required anymore. But I don't know.

xermicus: CC @athei

athei: With immutables being a thing now does it even make sense to support delegate calling by code hash? We probably just want to remove it and just allow by contract address.

xermicus: Yeah the nature of immutables turns delegate calling by hash into a potential footgun. I'm in favor of changing it to take an address instead, it's not required for Solidity. Should it for some reason turn out super viable for other languages we can still add it later.

crates/llvm-context/src/polkavm/evm/call.rs
ermalkaleci: I think we should replace weight (ref_time, proof_size) with just gas on revive. Solidity has no concept of PoV and all weight (substrate) related operations should be handled by pallet-revive (gas to weight and so on)

athei: It cannot really be handled by pallet-revive alone. Since Solidity has no concept of it we have to assume some default limits for proof size and storage deposit limit. As of right now the default is just "use all of the remaining". This is the most compatible but doesn't allow proper limiting. The plan is to expose the extended interface via pre-compile to allow Solidity contracts to limit all dimensions.

That being said, we decided to leave all parameters on the host function and revive just fills them out with `0`. This way other languages can at least use them.

xermicus: Yes you are right, for Solidity, ideally we just pass through what we get and let the pallet handle it. For other languages however this might be quite beneficial since they can just decide to not ignore it. Mapping everything it into the 1D (ref_time) space implies wasting block-space. Which is especially undesired for other languages like Rust/ink! because we expect people to use those for implementing more compute heavy logic. So to me it seems that ideally we'd have both. Or, since this is a pointer anyways, it could point to something like `enum Weights { Gas(U256), Native { ref_time: u64, proof_size: u64 } }`

xermicus: @ermalkaleci FWIW, in this PR it's fine to set it to remove the argument and set it to 0. The whole EVM gas vs. substrate weights problem is it's own topic.

athei: Of course we can provide some additional API that just takes "gas". But this would just be optimization. It has exactly the same semantics as filling `0` as we do now.

ermalkaleci: What if we replace `ref_time` with `gas` and let runtime do conversion to Weight (`proof_size` is optional, unused for solidity). Other contracts build without revive using Weight will have to convert `ref_time` to `gas`. At the end PolkaVM knows how to handle gas. `proof_size` remains optional, probably rename it to `storage_limit`

ermalkaleci: I don't see how `ref_time` can be equal to `gas`. 1 instruction means 1 unit of gas but it doesn't mean it's 1 picosecond. For brand new chains not depending on benchmarked `ref_time` it's easy because you can can measure how much gas is consumed into 1 seconds and use that reference to limit gas per block, but in case of parachains they need to convert between `gas` and `ref_time`

ermalkaleci: I propose we rename `ref_time` -> `gas_limit` and `proof_size` -> `storage_limit`

athei: > I don't see how ref_time can be equal to gas. 

It is just how we define gas. This is where we incompatible to EVM. We use different gas values to allow making certain operations cheaper than others.

ermalkaleci: True but I don't see how we can match 1 unit of gas with 1 unit of time (we may come close though). Anyway renaming to `gas_limit` and `storage_limit` makes more sense to me.

athei: > True but I don't see how we can match 1 unit of gas with 1 unit of time (we may come close though)

We are doing this: ` 1 gas = 1 ref_time = 1 picosecond of execution on the reference machine`

xermicus: Why renaming it?

xermicus: Ok I see but `gas_limit` and `storage_limit` are terms that neither exist in EVM nor in Substrate :) I think that this would only increase the confusion.

athei: And where do you want to rename them? Here in the recompiler they obviously don't come up?

ermalkaleci: Never mind. It's fine

crates/integration/contracts/Delegate.sol
ermalkaleci: just making sure emitted event belongs to caller and not callee

crates/integration/src/tests.rs
ermalkaleci: will add more cases

xermicus:
@ermalkaleci we will fix the CI first

xermicus:
@ermalkaleci merging main will fix the CI issue

crates/llvm-context/src/polkavm/evm/call.rs
```diff
@@ -60,7 +59,7 @@ where
     let arguments = &[
         flags.as_basic_value_enum(),
         address_pointer.value.as_basic_value_enum(),
-        context.integer_const(64, 0).as_basic_value_enum(),
+        gas.as_basic_value_enum(),
```
xermicus, line 62: I'm afraid this needs to be set to 0 (effectively ignoring it) until we figure out how to handle gas limits. I got some ideas but this is a later PR.

xermicus (approved):
Thank you @ermalkaleci!

PR #81, opened 2024-10-13
improve runtime api return value names

ermalkaleci:
Some change to runtime api call method to make it easy to read `ll` generated code.

before
```ll
%"runtime API call call" = call i32 @call(i32 %call_argument_pointer)
%is_success.not = icmp eq i32 %"runtime API call call", 0
```
after
```ll
%runtime_api_call_return_value = call i32 @call(i32 %call_argument_pointer)
%is_success = icmp eq i32 %runtime_api_call_return_value, 0
```


xermicus (approved):
Thanks!

PR #82, opened 2024-10-14
revive-runner: consider only non-reverted transactions as success

xermicus:
The pallet returns the `Ok` variant when the transaction did revert, so we should also consider the revert flag.


wpt967 (approved):
LGTM

PR #90, opened 2024-10-17
Makefile: remove duplicate workspace tests from the test target

xermicus:
Also moved the clippy target around next to other test related targets


wpt967 (approved):
LGTM

Issue #39, opened 2024-09-02
Not supported evm.bytecode and evm.assembly options in compiler configuration


smiasojed:
Although the tool claims that it does not support `evm.bytecode` and evm.assembly, it still returns it as part of the compilation output.

Used compiler config:
```
{
	"language": "Solidity",
	"settings": {
		"optimizer": {
			"enabled": true,
			"runs": 200
		},
		"outputSelection": {
			"*": {
			"": ["ast"],
			"*": ["abi", "metadata", "devdoc", "userdoc", "storageLayout", "evm.bytecode", "evm.legacyAssembly", "evm.deployedBytecode", "evm.methodIdentifiers", "evm.assembly"]
			}
		}
	}
}

```

It is expected that tool accepts `evm.bytecode` and `evm.assembly`

smiasojed:
`evm.gasEstimates` is also not supported

xermicus:
Stemming from an internal discussion:

- Provide what was requested; however
- Provide default values for incompatible data (if requested)

We substitute any byte-code incompatible fields with empty objects. This won't crash any tools relying on it, they can still request it but will get a zero or empty value. Which they can do nothing with so we should be save. Also this will make remix happy.

athei:
Both of them should be advertised as being available in order to maximize compatibility.

`evm.bytecode` should be returned but empty (we only have init code). `evm.assembly` should contain the PolkaVM dissassembly.

xermicus:
To me it'd also make sense to just provide the blob for both, the deploy code and runtime code? Because it is the same, it seems like we can just provide both for maximum compatibility.

smiasojed:
Remix uses `evm.bytecode` internally, so I think it is better to keep `bytecode` in sync with `deployedBytecode`.

athei:
Because of course they do. They depend on every little detail :(

smiasojed:
> `evm.assembly` should contain the PolkaVM dissassembly.

I think that we need to update polkavm-disassembler to be compatible with evm.legacyAssembly
```
"legacyAssembly": 
				{
							".code": [
								{
									"begin": 199,
									"end": 555,
									"name": "PUSH",
									"source": 0,
									"value": "80"
								}
                                                           ]
				}
```

xermicus:
Legacy assembly is not to confuse with EVM bytecode. Legacy assembly is technically speaking on of our IRs. It can be passed through as-is.

xermicus:
[...]

PR #95, opened 2024-10-23
Set evm.deployedBytecode to the value of evm.bytecode

smiasojed:
Related to #39 
Sync `evm.deployedBytecode` with `evm.bytecode`
Add evm.assembly flag support


crates/solidity/src/build/contract.rs
```diff
@@ -139,6 +139,9 @@ impl Contract {
         let bytecode = hex::encode(self.build.bytecode.as_slice());
         if let Some(evm) = standard_json_contract.evm.as_mut() {
             evm.modify(assembly_text, bytecode);
+            if standard_json_contract.ir_optimized.is_some() {
+                evm.assembly = None;
```
smiasojed, line 143: @xermicus I am not sure this, can I assume if ir_optimized is set, the legacyassembly as ir is not needed?

xermicus: ir-optimized is just another output option. They are two different things.

smiasojed: OK, removed

crates/solidity/src/solc/standard_json/output/contract/evm/mod.rs
```diff
@@ -29,7 +29,9 @@ pub struct EVM {
     /// Is reset by that of PolkaVM before yielding the compiled project artifacts.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub bytecode: Option<Bytecode>,
-    /// The contract deployed bytecode.
+    /// The deployed bytecode of the contract.
+    /// It is overwritten by PolkaVM before yielding the compiled project artifacts,
```
xermicus, line 33:
```suggestion
    /// It is overwritten with the PolkaVM blob before yielding the compiled project artifacts.
```

crates/solidity/src/solc/standard_json/output/contract/evm/mod.rs
```diff
@@ -29,7 +29,9 @@ pub struct EVM {
...
+    /// and set to the value of the contract's bytecode.
```
xermicus, line 34:
```suggestion
    /// Hence it will be the same as the runtime code but we keep both for compatibility reasons.
```

xermicus (approved):

Issue #93, opened 2024-10-18
`resolc` binary releases


xermicus:
We need to release binary version of the `resolc` Solidity frontend executable for Linux, MacOS and Windows.

Since there are many ways people can and do manage their `solc` binaries (e.g. download from GH release, use you distro package manager, npm install something, ...), I think our best bet is to have binary releases so tooling can support it.

For the linux binary see #45

athei:
What we should release are the following artifacts:

### For Production
- docker container with musl statically linked `resolc` and a fixed `solc` version in it 
    - We might later ship multiple versions of solc in this container

### For Dev
- `resolc` for x64 linux (same binary as in the docker)
- `resolc` universal binary for macOS
- `resolc` for 64bit windows

xermicus:
This is how zksolc [builds LLVM](https://github.com/matter-labs/era-compiler-llvm-builder). This should be close to what we need to do. We can also have a different repository with dedicated workflows for LLVM build releases.

xermicus:
Update: Depends on #86

Regarding the alpine release, see the example `Dockerfile` introduced in #154.

xermicus:
@athei correct me if I'm wrong but since the requirement for that Docker image aren't exactly clear yet, for now we will release:
- Linux (GNU and MUSL)
- Windows
- MacOS (universal)
- Wasm32 (emscripten)

mutantcornholio:
And as for LLVM, we need: 
* linux-gnu
* macos-gnu
* windows-gnu
* emscripten
* musl

athei:
@xermicus  Yes correct. Let's skip docker for now.

@mutantcornholio I don't understand your list. Why should it be any different from what Cyrill just listed?

> linux-gnu

Didn't we just agree that we use only the musl clib on Linux?

> macos-gnu

There is no such thing AFAIK.

> windows-gnu

You mean mingw? Why not use the msvc toolchain?

> musl

Thats just a name of a clib.

xermicus:
> [@xermicus](https://github.com/xermicus) Yes correct. Let's skip docker for now.
> 
> [@mutantcornholio](https://github.com/mutantcornholio) I don't understand your list. Why should it be any different from what Cyrill just listed?
> 
> > linux-gnu
> 
> Didn't we just agree that we use only the musl clib on Linux?
> 
> > macos-gnu
> 
> There is no such thing AFAIK.
> 
> > windows-gnu
> 
> You mean mingw? Why not use the msvc toolchain?
[...]

Issue #45, opened 2024-09-12
Create `alpine` release of revive compiler (musl.libc)


smiasojed:
Create the release workflow for `revive` compiler.
As artefacts provide:
1. revive-musl version
2. alpine docker
3. ~~revive-glibc version~~

xermicus:
The musl build should be beneficial for the remix backend.

athei:
Glibc version should not be necessary. A statically linked revive-musl should have zero dependencies.

athei:
@wpt967 Can you take care of this and then go back to the debugging work?

wpt967:
Sure.

xermicus:
@wpt967 thanks a lot!

I think how we should do it for now: The musl workflow should only run on pushes to the main branch and depend on the success of [build-ubuntu-x86](https://github.com/paritytech/revive/blob/main/.github/workflows/rust.yml#L13) (see [here](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run)). This second workflow can then just build musl, build revive linked against musl, and finally build and publish an alpine Docker image containing the musl revive build.

Should probably also just add a Makefile target install-musl that produces a release build of revive linked against musl.

So, essentially each time we push something to main and the CI passes we'll get a fresh image, which should be good to iterate quickly. Later we can think how to do releases properly (we also want some matrix testing against `solc` versions etc., but thats for later).

smiasojed:
In the Docker image, we need to include both 'resolc-musl' and 'solc-musl'. I would like to ask you to create the first revive release, which I can use as the base for the revive-remix-backend.

athei:
I wouldn't add the `musl` suffix to `resolc`. We should not do any dynamically linked releases anyways as they wouldn't be portable. Or is there any reason to do so?

smiasojed:
I just wanted to emphasize that we need both in the musl versions. We shouldn't have any suffix.

xermicus:
https://gitlab.com/taricorp/llvm-sys.rs/-/issues/44

With this we can give it at shot again. Just need to make sure we pull in the latest version through `inkwell`. Probably our best bet is on using https://github.com/rust-cross/rust-musl-cross as cross-compilation ought to work even with proc-macros.

athei:
So there is an issue with llvm-system which is fixed on master but no release, yet?

xermicus:
[...]

PR #97, opened 2024-10-25
Release CI

AndWeHaveAPlan:
cc https://github.com/paritytech/ci_cd/issues/1068

### DIFF
--- a/.github/Dockerfile-llvm
+++ b/.github/Dockerfile-llvm
@@ -1,0 +1,14 @@
+FROM alpine:3.20.3
+
+ARG RUST_VERSION=stable
+
+RUN apk add bash git cmake make g++ ninja python3 ncurses-static curl lld-dev lld lld-libs libc++-static
+RUN ninja --version
+
+RUN bash -c ./build-llvm.sh
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain ${RUST_VERSION}
+ENV PATH=/root/.cargo/bin:/llvm18.0/bin:${PATH}
+
+WORKDIR /app
+RUN REVIVE_INSTALL_DIR=$(pwd)/target/release
--- a/.github/Dockerfile-llvm
+++ b/.github/Dockerfile-llvm
@@ -5,6 +5,8 @@ ARG RUST_VERSION=stable
 RUN apk add bash git cmake make g++ ninja python3 ncurses-static curl lld-dev lld lld-libs libc++-static
 RUN ninja --version
 
+COPY build-llvm.sh .
+
 RUN bash -c ./build-llvm.sh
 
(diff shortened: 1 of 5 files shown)


.github/Dockerfile-llvm
```diff
@@ -0,0 +1,16 @@
+FROM alpine:3.20.3
+
+ARG RUST_VERSION=stable
+
+RUN apk add bash git cmake make g++ ninja python3 ncurses-static curl lld-dev lld lld-libs libc++-static
```
athei, line 5:
```suggestion
RUN apk add bash git cmake make ninja python3 ncurses-static curl
```

We should use the compilers from our custom built llvm. Especially gcc we don't want so we don't use it by accident.

.github/Dockerfile-llvm
```diff
@@ -0,0 +1,16 @@
+FROM alpine:3.20.3
+
+ARG RUST_VERSION=stable
+
+RUN apk add bash git cmake make g++ ninja python3 ncurses-static curl lld-dev lld lld-libs libc++-static
+RUN ninja --version
+
+COPY build-llvm.sh .
+
+RUN bash -c ./build-llvm.sh
```
athei, line 10:
```suggestion
RUN bash -c ./build-llvm.sh
```

We probably need to modify the `build-llvm.sh` file to use other flags. AFAIK @xermicus discovered that some other flags are needed to it statically links against LLVMs `libc++` instead of gccs  `libstdc++`.

xermicus: Yes, when I tried cross compiling to musl using the current build script, resolc was still linked against `libstdc++` (and segfaulted with an LLVM panic message). See also this [comment in the related issue](https://github.com/paritytech/revive/issues/93#issuecomment-2423776917). 

@wpt967 mentioned he was looking into it, what is the status?

athei: I told @wpt967 to de-pritotize this one since we are only supporting compilation REMIX for now. Until we have proper releases.

.github/Dockerfile-llvm
xermicus: The custom LLVM needs a bootstrapping compiler. I think we only want the RISC-V target in the custom LLVM build we link into resolc. If we need to provide a bootstrapping LLVM build for our supported platforms because the provided ones don't work, we should set up proper multistage LLVM builds.

xermicus (commented):
You will also need to change the `inkwell` dependency to use [my branch here](https://github.com/TheDan64/inkwell/pull/549) (why is explained in the [related issue](https://github.com/paritytech/revive/issues/45#issuecomment-2423171576)).

xermicus:
> You will also need to change the `inkwell` dependency to use [my branch here](https://github.com/TheDan64/inkwell/pull/549) (why is explained in the [related issue](https://github.com/paritytech/revive/issues/45#issuecomment-2423171576)).

This is now no longer needed.

xermicus:
Closing this in favor of #201

PR #100, opened 2024-10-25
support full storage key space

ermalkaleci:
This PR addresses two key issues:
- The storage pointer values will no longer be truncated to the register size, allowing for the use of arbitrary storage keys, not just those that fit within the register limits.
- Failed storage value reads will now guarantee to return the zero value on error

### DIFF
--- a/crates/integration/contracts/ERC20.sol
+++ b/crates/integration/contracts/ERC20.sol
@@ -4,9 +4,80 @@ pragma solidity ^0.8;
 
 /* runner.json
 {
+    "differential": true,
     "actions": [
     {
-      "Instantiate": {}
+        "Instantiate": {
+            "code": {
+                "Solidity": {
+                    "contract": "ERC20"
+                }
+            }
+        }
+    },
+    {
+        "Call": {
+            "dest": {
+                "Instantiated": 0
+            },
+            "data": "313ce567"
+        }
+    },
+    {
+        "Call": {
+            "dest": {
+                "Instantiated": 0
+            },
+            "data": "06fdde03"
+        }
+    },
+    {
+        "Call": {
+            "dest": {
+                "Instantiated": 0
+            },
+            "data": "95d89b41"
+        }
+    },
+    {
+        "Call": {
+            "dest": {
+                "Instantiated": 0
+            },
+            "data": "a0712d680000000000000000000000000000000000000000000000000000000000003039"
+        }
+    },
+    {
+        "Call": {
+            "dest": {
+                "Instantiated": 0
+            },
+            "data": "70a082310000000000000000000000000101010101010101010101010101010101010101"
+        }
+    },
+    {
+        "Call": {
+            "dest": {
+                "Instantiated": 0
+            },
+            "data": "a9059cbb000000000000000000000000ed27012c24fda47a661de241c4030ecb9d18a76d000000000000000000000000000000000000000000000000000000000000007b"
+        }
+    },
+    {
+        "Call": {
+            "dest": {
+                "Instantiated": 0
+            },
+            "data": "095ea7b3000000000000000000000000ed27012c24fda47a661de241c4030ecb9d18a76d00000000000000000000000000000000000000000000000000000000000000ff"
+        }
+    },
+    {
+        "Call": {
+            "dest": {
+                "Instantiated": 0
+            },
+            "data": "dd62ed3e0000000000000000000000000101010101010101010101010101010101010101000000000000000000000000ed27012c24fda47a661de241c4030ecb9d18a76d"
+        }
     }
   ]
 }
--- a/crates/integration/contracts/ERC20.sol
+++ b/crates/integration/contracts/ERC20.sol
@@ -6,80 +6,25 @@ pragma solidity ^0.8;
 {
     "differential": true,
     "actions": [
-    {
-        "Instantiate": {
-            "code": {
-                "Solidity": {
-                    "contract": "ERC20"
+        {
+            "Upload": {
+                "code": {
+                    "Solidity": {
+                        "contract": "ERC20"
+                    }
+                }
+            }
+        },
+        {
+            "Instantiate": {
+                "code": {
+                    "Solidity": {
+                        "contract": "ERC20Tester"
+                    }
                 }
             }
         }
-    },
-    {
-        "Call": {
-            "dest": {
-                "Instantiated": 0
-            },
-            "data": "313ce567"
-        }
-    },
-    {
-        "Call": {
-            "dest": {
-                "Instantiated": 0
-            },
-            "data": "06fdde03"
-        }
-    },
-    {
-        "Call": {
-            "dest": {
-                "Instantiated": 0
-            },
-            "data": "95d89b41"
-        }
-    },
-    {
-        "Call": {
-            "dest": {
-                "Instantiated": 0
-            },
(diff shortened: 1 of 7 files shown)


crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -669,20 +669,6 @@ where
...
-                let storage_key_pointer_casted = self.builder().build_ptr_to_int(
-                    storage_key_pointer.value,
-                    self.xlen_type(),
-                    "storage_key_pointer_casted",
-                )?;
-                self.builder()
-                    .build_store(storage_key_pointer.value, storage_key_value)?;
-
                 let storage_value_pointer =
                     self.build_alloca(self.word_type(), "storage_value_pointer");
```
xermicus, line 673:
```suggestion
                    self.build_alloca(self.word_type(), "storage_value_pointer");
                    self.build_store(storage_value_pointer, self.word_const(0))?;
```
It looks to me like this instead of emiting the branch later on would be more efficient code size wise. Can you try this and revert the change below that emits the branch to confirm less code size increase for the ERC20 and flipper contracts In the diff above?

crates/llvm-context/src/polkavm/context/pointer.rs
```diff
@@ -76,6 +76,35 @@ impl<'ctx> Pointer<'ctx> {
         Self::new(r#type, address_space, value)
     }
 
+    /// Creates a new pointer with the specified `offset` for storage operations.
+    pub fn new_with_store_offset<D, T>(
```
xermicus, line 80: The right place for this would be `fn build_storage_pointer_argument_store` on context, similar to the other helpers.

crates/llvm-context/src/polkavm/context/pointer.rs
```diff
@@ -76,6 +76,35 @@ impl<'ctx> Pointer<'ctx> {
...
+        context: &Context<'ctx, D>,
+        address_space: AddressSpace,
+        r#type: T,
```
xermicus, line 83: This is not needed. We always return a register sized (xlen) pointer (because it contains the allocation pointer and no the storage pointer).

crates/llvm-context/src/polkavm/context/pointer.rs
```diff
@@ -76,6 +76,35 @@ impl<'ctx> Pointer<'ctx> {
...
+    ) -> Self
+    where
+        D: Dependency + Clone,
+        T: BasicType<'ctx>,
+    {
+        assert_ne!(
+            address_space,
+            AddressSpace::Stack,
+            "Stack pointers cannot be addressed"
+        );
```
xermicus, line 95: This should rather assert that the address space is either Storage or TransientStorage?

crates/llvm-context/src/polkavm/context/pointer.rs
```diff
@@ -76,6 +76,35 @@ impl<'ctx> Pointer<'ctx> {
...
+
+        // TODO: should we swap the key???
+        // let swapped = context.build_byte_swap(offset.into()).unwrap();
```
xermicus, line 98: No. It would fail any differential test against EVM, wouldn't it?

xermicus (commented):
Very nice catch, thanks!

So this PR fixes two problems: The storage pointer values should not be truncated to the register size so that arbitrary storage keys work, not just ones that fall below our register size. And failed storage value reads now guarantee to return the zero value on error. Could you please add that to the description?

xermicus:
Ah okay, of course the tip bot isn't set up for this repo

crates/llvm-context/src/polkavm/context/pointer.rs
ermalkaleci: We're currently  reversing key on specs to make it work. It may be useful for readability on db side i.e doing query but will add more instructions. So I am not sure we need it

ermalkaleci:
> Very nice catch, thanks!
> 
> So this PR fixes two problems: The storage pointer values should not be truncated to the register size so that arbitrary storage keys work, not just ones that fall below our register size. And failed storage value reads now guarantee to return the zero value on error. Could you please add that to the description?

yes sure, I was waiting for you review first :)

crates/llvm-context/src/polkavm/context/mod.rs
xermicus: Thinking about it, adding that to the Storage.sol test fixture (reading a nonexistant key and asserting it returns a zero value) wouldn't harm.

crates/integration/contracts/ERC20.sol
xermicus, line 1: Would you mind using an instantiator contract to test this instead? It will make it a lot more obvious what is going on. And also be much easier to port into the upcoming new test format.

The drawback is that it's harder to spot what exactly failed. But for one this won't be an issue anymore once we implement execution tracing. And I mainly added this fixture to observe code size, we don't want to test a single specific functionality with it, so it doesn't matter as much. 

Appreciate the increased test coverage here, please forgive me this nitpick

ermalkaleci: you mean like we do with ext_code_hash? https://github.com/paritytech/revive/blob/cc38c374810f37a6d0a89b46681e944050a8d4d0/crates/integration/src/tests.rs#L238

xermicus: No, like this: https://github.com/paritytech/revive/blob/main/crates/integration/contracts/Immutables.sol#L37

ermalkaleci: oh, got it

crates/llvm-context/src/polkavm/context/mod.rs
xermicus: We will definitively check again when we have 64bit support, gas metering and everything. Also if either one is more efficient in either dimension (code size vs. execution performance), we can just emit either one based on the optimizations request (i.e. size optimization will emit this but not -O3).

xermicus: But by default we should optimize for size here. Storage operations are costly anyways.

crates/integration/contracts/ERC20.sol
ermalkaleci: I had to hardcode code_size since it's not implemented on pallet-revive

xermicus: Yeah that's fine. A pallet PR for implementing it is already in the pipeline.

ermalkaleci: I figured out a way to avoid generating extcodesize

crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -698,17 +686,13 @@ where
...
                         self.xlen_type().const_int(transient as u64, false).into(),
-                        storage_key_pointer_casted.into(),
+                        pointer.to_int(self).into(),
                         self.xlen_type().const_all_ones().into(),
                         storage_value_pointer.to_int(self).into(),
                         storage_value_length_pointer.to_int(self).into(),
                     ],
                 );
 
-                // We do not to check the return value.
```
xermicus, line 708: The comment is still valid though, could we leave it in place here?

xermicus (approved):

crates/llvm-context/src/polkavm/context/mod.rs
ermalkaleci: was removed when I added the check 😅

PR #105, opened 2024-10-30
display the git revision version information

xermicus:
Looks like this:

```
❯ resolc --version
Solidity frontend for the revive compiler v0.1.0 rev-36d9317 (LLVM build (18, 1, 8))
```

### DIFF
--- a/crates/solidity/build.rs
+++ b/crates/solidity/build.rs
@@ -1,0 +1,9 @@
+fn main() {
+    let git_rev = std::process::Command::new("git")
+        .args(["rev-parse", "--short", "HEAD"])
+        .output()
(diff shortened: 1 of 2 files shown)


smiasojed:
could we align in with the `solc` version format?
0.1.0-dev+commit.36d9317

xermicus:
Why? This isn't supposed to be machine readable and we also want to display the LLVM build used.

Changed it to append `+commit.rev`  as you suggested.

smiasojed (approved):

smiasojed:
OK, It is fine to have llvm version there.
Please consider updating contract metadata with correct compiler version

xermicus:
>  Please consider updating contract metadata with correct compiler version

Yep, will do that too, thanks

PR #109, opened 2024-10-31
use the long revive version string in the contract metadata

### LINKED ISSUE #50: Rename fields in contracts metadata
In the metadata we provide following informations:
```
						"solc_zkvm_edition": null,
						"zk_version": "0.1.0"
```
I suppose we should change it to `revive_version`/`resolc_version`

xermicus:
Closes #50

### DIFF
--- a/crates/solidity/src/project/contract/metadata.rs
+++ b/crates/solidity/src/project/contract/metadata.rs
@@ -2,6 +2,8 @@
 
 use serde::Serialize;
 
+use crate::ResolcVersion;
+
 /// The Solidity contract metadata.
 /// Is used to append the metadata hash to the contract bytecode.
 #[derive(Debug, Serialize)]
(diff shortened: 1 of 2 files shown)


crates/solidity/src/project/contract/metadata.rs
```diff
@@ -24,14 +26,13 @@ impl Metadata {
         solc_metadata: serde_json::Value,
         solc_version: semver::Version,
         revive_pallet_version: Option<semver::Version>,
-        revive_version: semver::Version,
         optimizer_settings: revive_llvm_context::OptimizerSettings,
     ) -> Self {
         Self {
             solc_metadata,
             solc_version,
```
smiasojed, line 33: maybe `solc_version` should be also in long format?

smiasojed:
Maybe we should have `-dev-5` in resolc --version output:
`Solidity frontend for the revive compiler version 0.1.0-dev-5+commit.e2e0270.llvm-18.1`

smiasojed (approved):

crates/solidity/src/project/contract/metadata.rs
xermicus: Ok, changed it

Issue #98, opened 2024-10-25
Fix automatic import resultion


xermicus:
[See here](https://github.com/matter-labs/era-compiler-solidity/issues/98) for a detailed write up.

davidk-pt:
I can take this, although, I'll be off since Dec 17 and come back after new year. If no one else will work on this in the meantime I can assign to myself

davidk-pt:
I won't be able to finish this as I'm leaving parity, sorry

xermicus:
Should ideally be done after #91

PR #146, opened 2025-01-06
Adds import callback resolution to resolc compiler

davidk-pt:
Resolves https://github.com/paritytech/revive/issues/98

Let me know if this looks good and I'll write NodeJS tests.

Solidity compiler exports all sources, so I just add those in our source code files.

Tested with these two files:
`test.sol`
```
// SPDX-License-Identifier: MIT

pragma solidity >=0.4.16;

import "./callable.sol";

contract Main {
    function main(Callable callable) public returns(uint) {
        return callable.f(5);
    }
}
```

`callable.sol`
```
// SPDX-License-Identifier: MIT

pragma solidity >=0.4.16;

contract Callable {
    function f(uint a) public pure returns(uint) {
        return a * 2;
    }
}
```
This is the output
```
➜  revive git:(davidk/add-import-resolution-to-resolc) ✗ resolc ./test.sol
Warning: Function state mutability can be restricted to pure
 --> ./test.sol:8:5:
  |
8 |     function main(Callable callable) public returns(uint) {
  |     ^ (Relevant source part starts here and spans across multiple lines).


Compiler run successful. No output requested. Use --asm and --bin flags.
```

### DIFF
--- a/crates/solidity/src/lib.rs
+++ b/crates/solidity/src/lib.rs
@@ -52,7 +52,7 @@ pub use self::warning::Warning;
 pub mod test_utils;
 pub mod tests;
 
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet};
 use std::path::PathBuf;
 
 /// Runs the Yul mode.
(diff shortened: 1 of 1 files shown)


xermicus (commented):
Thanks for tackling this. The PR referenced from matter labs fixes this in `fn standard_json` instead. Why did you fix it differently?

The matter-labs PR for this also seems much more complete. Don't we need to backport all of those changes too?

PR #151, opened 2025-01-10
multiple resolc fixes and improvements

Brianspha:
- Error out early on compiler invocations with invalid base path or include path flags.
- Do not error out if no files and no errors were produced. This aligns `resolc` closer to `sloc`. 
- Add a CLI test with an involved fixture containing multiple contract remappings to properly testing the standard JSON path.
- Fixes input normalization in the Wasm version.


crates/solidity/src/solc/solc_compiler.rs
```diff
@@ -52,46 +64,73 @@ impl Compiler for SolcCompiler {
...
             anyhow::anyhow!("{} subprocess spawning error: {:?}", self.executable, error)
         })?;
-        process
-            .stdin
-            .as_ref()
-            .ok_or_else(|| anyhow::anyhow!("{} stdin getting error", self.executable))?
-            .write_all(input_json.as_slice())
-            .map_err(|error| {
+
+        let suppressed_warnings = input.suppressed_warnings.take().unwrap_or_default();
```
smiasojed, line 111: Please keep in sync changes in 
 `solc_compiler.rs` with https://github.com/paritytech/revive/blob/main/crates/solidity/src/solc/soljson_compiler.rs

crates/solidity/Cargo.toml
```diff
@@ -40,7 +40,7 @@ inkwell = { workspace = true }
 
 revive-common = { workspace = true }
 revive-llvm-context = { workspace = true }
-
+tracing= "0.1.41"
```
xermicus, line 43: Why is this added?

Use `log` instead given it is only being used for logging which `log` already provides, given `log` is already a workspace dependency. Please use existing workspace dependencies wherever possible.

New dependencies should be added to the workspace dependencies but adding this is unnecessary.

crates/solidity/src/solc/solc_compiler.rs
```diff
@@ -52,46 +63,73 @@ impl Compiler for SolcCompiler {
         allow_paths: Option<String>,
     ) -> anyhow::Result<StandardJsonOutput> {
         let version = self.version()?;
+        trace!("In standard_json");
```
xermicus, line 66: Please remove all logging, it is unrelated to the bugfix this PR is addressing.

If you want to enable logging or tracing please do it in another PR, use the `log` crate and initialize the `env_logger` in the main function. make sure to log in all relevant places and use `

crates/solidity/src/solc/standard_json/output/mod.rs
```diff
@@ -58,21 +58,51 @@ impl Output {
...
+
+                        if has_errors {
+                            error!("Compiler errors found: {:#?}", errors);
+                            anyhow::bail!(
+                                "{}",
+                                serde_json::to_string_pretty(errors).expect("Always valid")
+                            );
+                        }
+
+                        warn!("Compiler warnings found: {:#?}", errors);
```
xermicus, line 91: Useless log statements as everything should already be displayed via the compiler warnings and errors.

xermicus (changes requested):
Thanks for tackling this! Can you please add a test to verify the bugfix?

crates/solidity/Cargo.toml
Brianspha: The test will require modification of `standard_json` in lib.rs to be modified to not read from std hence i didnt include it here is that still fine?

crates/solidity/src/solc/solc_compiler.rs
Brianspha: Explain @smiasojed

crates/solidity/Cargo.toml
Brianspha: Logs are just for testing i wasnt sure what the rules were around logging but i will remove them.

xermicus: >  The test will require modification of standard_json in lib.rs to be modified to not read from std

Why?

> Logs are just for testing i wasnt sure what the rules were around logging but i will remove them.

Thanks

Brianspha: @xermicus i wasnt able to get reading and writing to std in a test environment with out getting the BROKEN PIPE error or the test hanging since standard_json calls process.exit(0) which meant that reading the results from compilation will never execute

Brianspha: Heres a code share of the file i used locally to test
https://codeshare.io/VN0Pl3 ignore any unused imports i was trying different solutions to get reading and writting to std but failed so had to take the easy route of modifying standard_json

Brianspha: test file: can be found on this codeshare ensure the file is in the root of the solidity crate https://codeshare.io/DANvm3 @xermicus

Brianspha: Alternatively i can push the local branch i created for testing then we can delete it after you verify if everything is ok?

crates/solidity/src/solc/solc_compiler.rs
smiasojed: First of all you can build revive as native binary or wasm blob, which you can use in browsers.
solc_compiler is used in native compiler binary and soljson_compiler in wasm version of compiler. If you introduce change to solc_compiler only, it means that resolc binary will not generate the same output as resolc.wasm compiler version

crates/solidity/Cargo.toml
Brianspha: Additionally i used a source file that is similar to what foundry generates

xermicus: It should be testable in a [CLI test](https://github.com/paritytech/revive/tree/main/crates/solidity/src/tests/cli-tests).

Brianspha: Aha i didnt focus on this part of the code do i need to add a test there?

xermicus: Yes. In the CLI tests you have complete control over the compiler invocation and configuration.

Brianspha: Alright i wasnt aware i needed to change that part of the code will do so asap

crates/solidity/src/solc/solc_compiler.rs
Brianspha: Alright but when you said above `Please keep in sync changes in` what does this mean?

smiasojed: I mean that any changes made to `solc_compiler.rs` should also be reflected in [`soljson_compiler.rs`](https://github.com/paritytech/revive/blob/main/crates/solidity/src/solc/soljson_compiler.rs)

Brianspha: @smiasojed please check again

crates/solidity/src/solc/solc_compiler.rs
```diff
@@ -1,16 +1,24 @@
...
+use std::path::PathBuf;
 
 use super::Compiler;
+// `--base-path` was introduced in 0.6.9 <https://github.com/ethereum/solidity/releases/tag/v0.6.9>
+pub static SUPPORTS_BASE_PATH: Lazy<VersionReq> =
+    Lazy::new(|| VersionReq::parse(">=0.6.9").unwrap());
+
+// `--include-path` was introduced in 0.8.8 <https://github.com/ethereum/solidity/releases/tag/v0.8.8>
+pub static SUPPORTS_INCLUDE_PATH: Lazy<VersionReq> =
+    Lazy::new(|| VersionReq::parse(">=0.8.8").unwrap());
```
xermicus, line 21: Please move those next to the other solc version defined constants in `crates/solidity/src/solc/mod.rs`.

Please also rename to follow the existing naming (`FIRST_BASE_PATH_VERSION`  and `FIRST_INCLUDE_PATH_VERSION`)

crates/solidity/src/solc/soljson_compiler.rs
```diff
@@ -35,20 +35,24 @@ impl Compiler for SoljsonCompiler {
         _allow_paths: Option<String>,
     ) -> anyhow::Result<StandardJsonOutput> {
         let version = self.version()?;
+        input.normalize(&version.default);
```
xermicus, line 38: Nice catch!

xermicus (commented):
The overall idea of aborting on errors and only pass valid flags seems correct, thanks!

crates/solidity/src/solc/solc_compiler.rs
```diff
@@ -56,42 +64,58 @@ impl Compiler for SolcCompiler {
...
+
+                if SUPPORTS_INCLUDE_PATH.matches(&version.default) {
+                    // `--base-path` and `--include-path` conflict if set to the same path, so
+                    // as a precaution, we ensure here that the `--base-path` is not also used
+                    // for `--include-path`
+                    for path in include_paths.iter().filter(|p| *p != base_path) {
+                        command.arg("--include-path").arg(Path::new(path));
+                    }
+                }
+            }
```
xermicus, line 80: I think instead of silently swallong the flag it should error out here. The user passing invalid solc arguments hints at a problem. We should not silenty ignore but treat it as an error instead.

crates/solidity/src/solc/solc_compiler.rs
```diff
@@ -56,42 +64,58 @@ impl Compiler for SolcCompiler {
...
-        }
-        for include_path in include_paths.into_iter() {
-            command.arg("--include-path");
-            command.arg(include_path);
+        if let Some(base_path) = &base_path {
+            if SUPPORTS_BASE_PATH.matches(&version.default) {
+                command.arg("--base-path").arg(base_path);
+
+                if SUPPORTS_INCLUDE_PATH.matches(&version.default) {
+                    // `--base-path` and `--include-path` conflict if set to the same path, so
```
xermicus, line 73: This comment is wrong.

crates/solidity/src/tests/cli-tests/tests/common.test.ts
```diff
@@ -142,3 +145,53 @@ describe("Run resolc with source debug information, check LLVM debug-info", () =
...
+
+        const input = JSON.parse(shell.cat(inputFile).toString());
+
+        Object.entries(input.sources).forEach(([sourcePath, source]: [string, any]) => {
+            const filePath = path.join(contractsDir, sourcePath);
+            shell.mkdir('-p', path.dirname(filePath));
+            shell.ShellString(source.content).to(filePath);
+            // I added this here to ensure that we can check for any files
+            // Not being written to temp dirs BUUUT i can remove it
+            console.log(`Wrote source file: ${sourcePath}`);
```
xermicus, line 166: Remove the comment and also the `console.log` statement if it is irrelevant for the test.

crates/solidity/src/tests/cli-tests/tests/common.test.ts
```diff
@@ -142,3 +145,53 @@ describe("Run resolc with source debug information, check LLVM debug-info", () =
...
+            result = executeCommand(command);
+
+            shell.rm(tempInputFile);
+
+        });
+
+        it("Compiler run successful", () => {
+            // Here we want to filter any warnings from the errors
+            // Respone since warnings arent really errors but exists
+            // As precautionary measures for devs
```
xermicus, line 192: This comment is very confusing.
```suggestion
        it("Compiler run successful without emiting warnings", () => {
```

crates/solidity/src/tests/cli-tests/tests/common.test.ts
```diff
@@ -142,3 +145,53 @@ describe("Run resolc with source debug information, check LLVM debug-info", () =
...
+            const parsedResults = JSON.parse(result.output)
+            expect(parsedResults.errors.filter((error: { type: string; }) => error.type != 'Warning')).toEqual([]);
```
xermicus, line 194: Why is this only checking for warnings?

If we expect errors, explicitly check against them. If no errors are expected, assert it or leave a comment on why it is unnecesary to do so.

crates/solidity/src/tests/cli-tests/tests/common.test.ts
```diff
@@ -142,3 +145,53 @@ describe("Run resolc with source debug information, check LLVM debug-info", () =
...
+    });
+
+    describe("Output with all path options", () => {
+        let result: { exitCode: number; output: string };
+
+        beforeAll(() => {
+            const tempInputFile = path.join(contractsDir, 'temp-input.json');
+            shell.cp(inputFile, tempInputFile);
+
+            const command = `cat "${tempInputFile}" | resolc --standard-json --base-path "${contractsDir}" --include-path "${contractsDir}" --allow-paths "${contractsDir}"`;
```
xermicus, line 181: This doesn't seem to be portable. I feel like it should setup STDIN via some shelljs API?

crates/solidity/src/solc/solc_compiler.rs
Brianspha: But remember this will be used by various tools with different solc versions so would we then need to abort the whole process of compilation

crates/solidity/src/tests/cli-tests/tests/common.test.ts
Brianspha: Remember errors= Warning,Error we only want to check for actual errors

xermicus: Nevermind, I misread the filter predicate. I was just confused by the comment above. It's fine like this, please just change the comment (sufficient to just say it is checking for no errors emitted).

crates/solidity/src/solc/solc_compiler.rs
xermicus: This is exactly my point. Invoking the compiler with invalid configuration should abort the compilation process instead of silently ignoring it.

Brianspha:
Due to changes introduced by the main branch (llvm config changes) i can no longer run any code changes locally as such i need to commit any changes and wait for them to fail/pass on the branch  @xermicus

xermicus:
Hi @Brianspha, sorry for the inconvenience. We slightly changed the way the LLVM dependency is resolved to fix some issues in the build process.

Please `export LLVM_SYS_181_PREFIX=/path/to/your/llvm/root/dir` or [try a clean build](https://github.com/paritytech/revive?tab=readme-ov-file#the-resolc-solidity-frontend). However rebuilding LLVM should not be necessary. The only thing changed is how the LLVM dependency is looked up.

xermicus (commented):
Looks good. Just some minor things to iron out.

crates/solidity/src/solc/solc_compiler.rs
```diff
@@ -1,17 +1,16 @@
...
-use std::path::PathBuf;
-
 use crate::solc::combined_json::CombinedJson;
 use crate::solc::pipeline::Pipeline;
 use crate::solc::standard_json::input::Input as StandardJsonInput;
 use crate::solc::standard_json::output::Output as StandardJsonOutput;
 use crate::solc::version::Version;
+use std::io::Write;
+use std::path::Path;
+use std::path::PathBuf;
```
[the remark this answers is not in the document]
Brianspha: Done

crates/solidity/src/solc/standard_json/output/mod.rs
```diff
@@ -64,15 +63,25 @@ impl Output {
...
+                        }
+                        BTreeMap::new()
+                    }
+                    None => {
+                        anyhow::bail!(
+                            "Unknown project assembling error - no contracts or errors present"
+                        );
+                    }
+                }
+            },
```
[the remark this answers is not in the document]
Brianspha: Good suggestion

xermicus (approved):

xermicus:
@Brianspha thanks a lot for your contribution!

Since the PR title and description it no longer matched the actual changes (it doesn't fix a remapping issue - the added test would also pass in `main`) I updated both to more accurately reflect what it does.

PR #154, opened 2025-01-11
revive llvm builder utility

xermicus:
Pre-eliminary support for LLVM releases and `resolc` binary releases by streamlining the build process for all supported hosts platforms.
- Introduce the `revive-llvm-builder` crate with the `revive-llvm` builder utilty. 
- Do not rely on the LLVM dependency in `$PATH` to decouple the system LLVM installation from the LLVM host dependency.
- Fix the emscripten build by decoupling the host and native LLVM dependencies. Thus allowing a single LLVM emscripten release that can be used on any host platform.
- An example Dockerfile building an alpine container with a fully statically linked `resolc` ELF binary.
- Remove the Debian builder utilities and workflow.


smiasojed:
@xermicus it seems that you forgot to remove `clone-llvm.sh` or is it still needed somewhere?

xermicus:
@smiasojed thanks for pointing, made a follow up to fix it: #159

PR #159, opened 2025-01-14
Fix the Wasm build cache

xermicus:
Follow-ups for #154 

- Fix the cache in the Wasm build
- Remove a no longer needed script
- The Wasm build job uses `parity-large`


.github/workflows/build-revive-wasm.yml
```diff
@@ -11,6 +11,7 @@ env:
   REVIVE_WASM_INSTALL_DIR: ${{ github.workspace }}/target/wasm32-unknown-emscripten/release
   EMSCRIPTEN_VERSION: 3.1.64
```
smiasojed, line 12: is used this emscripten version?

.github/workflows/build-revive-wasm.yml
```diff
@@ -11,6 +11,7 @@ env:
...
   BUN_VERSION: 1.1.43
+  REVIVE_LLVM_NATIVE_RELATIVE: target-llvm/gnu/target-final
```
smiasojed, line 14: what does it mean this `target-final`?

.github/workflows/build-revive-wasm.yml
```diff
@@ -35,21 +38,27 @@ jobs:
...
+          # Clone LLVM and install LLVM and the Emscripten SDK
           make install-llvm-builder
           revive-llvm --target-env emscripten clone
 
+      - name: Setup revive environment variable
+        run: |
+          echo "LLVM_SYS_181_PREFIX=$(pwd)/${REVIVE_LLVM_NATIVE_RELATIVE}" >> $GITHUB_ENV
+          echo "REVIVE_LLVM_TARGET_PREFIX=${PWD}/target-llvm/emscripten/target-final" >> $GITHUB_ENV
+          
+      - name: Build host LLVM
```
smiasojed, line 50:
```suggestion
      - name: Get host LLVM
```

.github/workflows/build-revive-wasm.yml
smiasojed, line 14: maybe HOST_LLVM_DIR?

.github/workflows/build-revive-wasm.yml
```diff
@@ -26,30 +25,43 @@ jobs:
...
+        with:
+          profile: minimal
+          toolchain: stable
+          components: rust-src
+          target: wasm32-unknown-emscripten
+
+      - name: Install LLVM build dependencies
+        run: |
           make install-llvm-builder
           revive-llvm --target-env emscripten clone
```
mutantcornholio, line 48: Shouldn't this also be guarded with `steps.cache-llvm.outputs.cache-hit != 'true'`?

xermicus: No; `revive-llvm --target-env emscripten clone` clones LLVM and the Emscripten SDK. The Emscripten SDK is required when compiling revive to Wasm.

xermicus: That being said, `revive-llvm-builder` could be more fine-granular in that regard. I.e. rework the `clone` command into a `dependency` command, which has more granular control over what it install. Then, during the LLVM build, we would install both, but during the revive build only emsdk.

But that's entirely optional and if desired should be done in another PR.

mutantcornholio: So we build compile revive-llvm just to clone Emscripten SDK?

xermicus: Yes. Note that it doesn't only clone but also install the right version.

xermicus: The `revive-llvm` utility itself could also be cached 🙂 But that should also be a follow-up; I want to merge this here ASAP to unblock pending PRs.

mutantcornholio (approved):

PR #164, opened 2025-01-15
llvm-builder testing workflow

mutantcornholio:


.github/workflows/revive-llvm-test.yml
xermicus, line 1: For the `revive-llvm` test it might be worthwhile to set `RUST_LOG=trace` too

crates/llvm-builder/tests/build.rs
```diff
@@ -53,6 +53,16 @@ fn clone_build_and_clean_musl() -> anyhow::Result<()> {
...
 
+    Command::cargo_bin(common::REVIVE_LLVM)?
+        .current_dir(test_dir.path())
+        .arg("build")
+        .arg("--llvm-projects")
+        .arg("clang")
+        .arg("--llvm-projects")
+        .arg("lld")
+        .assert()
+        .success();
```
xermicus, line 64: Marking it here but can you remove the `builtins` command below as a drive-by? The builtins are built anyways during the `build` command.

crates/llvm-builder/tests/build.rs
```diff
@@ -92,6 +102,16 @@ fn debug_build_with_tests_coverage() -> anyhow::Result<()> {
...
+
```
xermicus, line 114: Shouldn't this and the below command be merged? This here builds LLVM, clang and LLD. Below builds LLVM with tests and coverage.

crates/llvm-builder/tests/build.rs
```diff
@@ -116,6 +136,16 @@ fn build_with_sanitizers() -> anyhow::Result<()> {
         .assert()
         .success();
 
+    Command::cargo_bin(common::REVIVE_LLVM)?
```
xermicus, line 139: Same as above

crates/llvm-builder/tests/build.rs
```diff
@@ -134,9 +164,30 @@ fn clone_build_and_clean_emscripten() -> anyhow::Result<()> {
...
+        .current_dir(test_dir.path())
+        .arg("clone")
+        .assert()
+        .success();
+
+    Command::cargo_bin(common::REVIVE_LLVM)?
+        .current_dir(test_dir.path())
+        .arg("build")
+        .arg("--llvm-projects")
+        .arg("clang")
```
xermicus, line 178:
```suggestion
```

We should only need LLD; this will make the test consume less resources.

.github/workflows/revive-llvm-test.yml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+    types: [assigned, opened, synchronize, reopened]
+    paths:
+      - 'LLVM.lock'
+      - 'crates/llvm-builder/**'
+      - '.github/workflows/revive-llvm-test.yml'
+
+jobs:
+  test:
+    runs-on: parity-large
```
xermicus, line 11: We will release on Windows and MacOS too, so this probably should be a matrix test with those platforms to avoid any problems during releasing.

mutantcornholio:
Okay, taking this PR back to draft, let's see how win & mac will run the tests

crates/llvm-builder/tests/build.rs
mutantcornholio: That's host llvm, you probably meant the next command, right?

xermicus: Yes sorry the host LLVM needs clang and LLVM.

mutantcornholio:
Removed MacOS from the pipeline, these tests can't finish in 6h on public runners.

xermicus:
> Removed MacOS from the pipeline, these tests can't finish in 6h on public runners.

But they shouldn't be executed as they are now only run on linux? Or am I missing something?

mutantcornholio:
> > Removed MacOS from the pipeline, these tests can't finish in 6h on public runners.
> 
> But they shouldn't be executed as they are now only run on linux? Or am I missing something?

Yes, that's what I'm saying. I've added them in previous revision; they didn't work, so I've removed them, leaving only linux.

xermicus:
But now you only run the test on linux. It should run on all platforms.

xermicus:
What I'm saying: They shouldn't be executed on macos because they are gated to only exist on linux. Which in turns allows for a matrix with all of our supported platforms. I'd like to have those tests detect problems early for all platforms we are releasing, not just Linux.

mutantcornholio:
Okay, it's coming back to me now 😅

I think we didn't undetstand each other, as I was talking about the whole test suite, while you were meant only the now gated tests; while both of us were referring to them as "tests" or "they"

xermicus:
Yeah in the context of this PR with "tests" I mean the LLVM test only 🙂

.github/workflows/revive-llvm-test.yml
```diff
@@ -0,0 +1,44 @@
+on:
+  pull_request:
+    types: [assigned, opened, synchronize, reopened]
+    paths:
+      - 'LLVM.lock'
+      - 'crates/llvm-builder/**'
```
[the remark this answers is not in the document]
mutantcornholio: Why wouldn't we want to run these tests on changes to `crates/llvm-builder/Cargo.toml`?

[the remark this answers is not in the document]
xermicus: The LLVM builder is still checked for successful compilation and even clippy anyways. So the only thing which can break the LLVM builder are dependencies violating semver.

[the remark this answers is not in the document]
mutantcornholio: I'm just trying to understand which cases are you suggesting to exclude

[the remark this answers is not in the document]
xermicus: I'm not sure if I really follow, I mean what I suggest is to exclude anything that likely won't break the LLVM. Which is highly unlikely for `Cargo.toml` given:
- The dependencies are workspace dependencies anyways
- The crate is still being checked and lintend anyways

But it doesn't matter much anyways. I'll leave it up to you.

.github/workflows/revive-llvm-test.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: |
+          brew install ninja
+
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: stable
+          components: rust-src
+          target: wasm32-unknown-emscripten
+          rustflags: ""
+
```
xermicus, line 34: This should be unnecessary 

```suggestion
```

xermicus (approved):

.github/workflows/revive-llvm-test.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Install apt dependencies
+        if: matrix.runner == 'parity-large'
+        run: |
+          sudo apt update && sudo apt-get install -y cmake ninja-build curl git libssl-dev pkg-config clang lld musl
+
+      - name: Install macos dependencies
+        if: matrix.runner == 'macos-14' || matrix.runner == 'macos-13'
+        run: |
+          brew install ninja
+
```
xermicus, line 27: I wasn't aware that on the parity runner there is no rust installed at all so we should just remove the additional target

```suggestion

      - uses: actions-rust-lang/setup-rust-toolchain@v1
        with:
          toolchain: stable
          components: rust-src
          rustflags: ""

```

PR #168, opened 2025-01-16
llvm-builder: do not build clang by default

xermicus:
We only need LLD for cross compilation. This significantly reduces the LLVM build times in a cross compilation scenario. Update the README as a drive-by.


Makefile
```diff
@@ -36,7 +36,7 @@ install-llvm-builder:
 
 install-llvm: install-llvm-builder
 	revive-llvm clone
-	revive-llvm build
+	revive-llvm build --llvm-projects lld --llvm-projects clang
```
athei, line 39: Why do we need `clang` for the host LLVM? We only need it when we plan to use that LLVM to cross compile later?

xermicus: `clang` will be used in multiple `build.rs`  scripts. It compiles the runtime API and the `stdlib`. This approach decouples the revive LLVM host dependency completely from the system LLVM dependency.

athei: I see. Wasn't aware that we need to compile C code.

PR #169, opened 2025-01-17
Fix issue with relative path in web worker

smiasojed:
Remove reloading of the resolc.js file.


xermicus (approved):

xermicus:
I didn't expect to time it out especially with the bumped limits. Why is it so much worse if it's a single file?

athei:
Maybe just problems with very large text files?

smiasojed:
> Maybe just (memory) problems with very large text files?

Yes, it is a memory problem caused by the SINGLE_FILE option. It seems that this is implemented in a non-optimal way.

PR #173, opened 2025-01-23
Pack resolc.js and resolc.wasm into single file

smiasojed:
Pack resolc.js and resolc.wasm into single file
Format JS files
Fix issue with Module overwriting in preloaded js.


athei:
This way of packing it works better in Firefox than the way we tried before?

smiasojed:
> This way of packing it works better in Firefox than the way we tried before?

No is does not, I am doing still experiments

Makefile
```diff
@@ -28,8 +28,19 @@ install-bin:
 install-npm:
 	npm install && npm fund
 
+TARGET_PATH=target/wasm32-unknown-emscripten/release
+WASM_FILE=$(TARGET_PATH)/resolc.wasm
+JS_FILE=$(TARGET_PATH)/resolc.js
+OUTPUT_FILE=$(TARGET_PATH)/resolc_packed.js
```
xermicus, line 34: Nit: Can we rename to something more telling? They are as general as variable names can get..

E.g.

`TARGET_PATH` -> `CARGO_WASM32_EMSCRIPTEN_TARGET_DIR_PATH`
`WASM_FILE` -> `RESOLC_WASM_TARGET_PATH`
`JS_FILE`-> `RESOLC_JS_TARGET_PATH`
`OUTPUT_FILE` -> `RESOLC_JS_PACKED_TARGET_PATH`

xermicus (approved):

Makefile
smiasojed: sure, good idea

Makefile
```diff
@@ -28,8 +28,19 @@ install-bin:
...
+
 install-wasm: install-npm
 	cargo build --target wasm32-unknown-emscripten -p revive-solidity --release --no-default-features
+	@if [ ! -f $(WASM_FILE) ]; then echo "Error: Missing $(WASM_FILE)"; exit 1; fi
+	@if [ ! -f $(JS_FILE) ]; then echo "Error: Missing $(JS_FILE)"; exit 1; fi
```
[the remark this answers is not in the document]
smiasojed: It should not happen I can remove this check if you want

js/utils/mini-lz4.js
```diff
@@ -0,0 +1,118 @@
+function uncompress(source, uncompressedSize) {
+/*
+Source https://github.com/ethereum/solidity/blob/develop/scripts/ci/mini-lz4.js
```
smiasojed, line 3: @xermicus This file is taken from Solidity. Is the licensing compliant for use?

xermicus: I don't think so as `ethereum/solidity` appears to be GPL licensed.

smiasojed: Ok, so I will rewrite this small part, and we will simply base it on the original mini-lz4. Does that work for you?

xermicus: Sounds good, however I have a bunch of higher level questions related to this:
1. Why is this needed?
2. Why do they use a custom compression implementation (or is it also a custom algorithm)?
3. Why not just use a library?

smiasojed: 1. It is used for revive wasm compression.
2. it standard lz4 compression, but needs to embeded it in the resolc.js to decompress wasm on the fly
3. What do you mean? Do you want to embed npm lz4 package to resolc.js?

smiasojed: Alternative solution: https://github.com/paritytech/revive/pull/176

smiasojed:
Approach in the #176 PR will be merged.

PR #176, opened 2025-01-27
Add all resolc dependencies to resolc_web.js file

smiasojed:
Add all resolc dependencies to resolc_web.js file
Format the JS files
Fix the issue with module overwriting in the preloaded JS


xermicus:
> We could temporarily host compilers on revive-remix-backend, where CSP is supported.

Since we are bundling a fixed `solc` version, this is useful on the backend only anyways? If so I don't see a problem.

js/build.js
```diff
@@ -0,0 +1,53 @@
+const fs = require("fs");
+const path = require("path");
+const { minify } = require("terser");
+
+const SOLJSON_URI =
+  "https://binaries.soliditylang.org/wasm/soljson-v0.8.28+commit.7893614a.js";
```
xermicus, line 6: Are we planning to this dynamically inside remix in the future?

I think this should be done client side. In remix , the user can select the `solc` version there. Remix will download `resolc` and bundle it together. This way we can support all `solc` versions (since `resolc` is always compatible with any `solc`, the `resolc` version can be hardcoded to whatever is currently compatible with the node).

smiasojed: How it works in Remix: Remix lists the compiler files and loads the selected `soljson-0.8.28+commit.xxxx`. The file must follow this format because the `soljson-0.8.28` part is used internally, for example, in pragma checking.

Steps to bundle it on the Remix side:
Modify the compiler tab to browse two lists of compilers.
Store this information in Remix's internal objects.
Update the web worker logic and API to accept two compilers.
Modify the non worker compilation flow in remix.
Modify the test logic, as it also expects a single compiler.


This change will affect many files in Remix. I wanted to implement it, but first, I need to run Remix's unit and E2E tests. A side effect of this change is that any future Remix updates will take more time to integrate.

I think I could remove soljson dep from revive and add it during compiler list generation on backend

xermicus: We have two options here:
1. Implement this properly in remix and don't bundle anything together.
2. Bundle `resolc` + `solc`. The `solc` version bundled is the latest (supported) version. We declare support for other `solc` versions as out of scope. Honestly I don't think remix is being used for serious dApp development and deployments but rather as a playground for quick experimentation, demos and so forth.

The second option would also allow us to spend much less resources in optimizing the compiler for the in-browser environments. Honestly I think the second option would allow us to move forward fast and focus on more important issues.

smiasojed: We need to decide where to host the compilers if GitHub Pages does not support enabling CSP headers.
When creating the resolc_packed (bundle) release, we must upload the WASM file to a server that supports CSP.

smiasojed: I will put it in the Remix repo. No one else will use it, so it's safe to keep it there.

xermicus (approved):
LGTM

.github/workflows/build-revive-wasm.yml
```diff
@@ -87,6 +87,7 @@ jobs:
           path: |
             ${{ env.REVIVE_WASM_INSTALL_DIR }}/resolc.js
```
athei, line 88: Do we still need to release it? Does it have any worth over `resolc_packed.js`? How would you use it?

.github/workflows/build-revive-wasm.yml
```diff
@@ -87,6 +87,7 @@ jobs:
...
             ${{ env.REVIVE_WASM_INSTALL_DIR }}/resolc.wasm
+            ${{ env.REVIVE_WASM_INSTALL_DIR }}/resolc_packed.js
```
athei, line 90: Does the suffix `packed` make sense here? Given that this is not really packed but loads the wasm file (i.e not containing it).

.github/workflows/build-revive-wasm.yml
smiasojed: resolc.js makes sense for use with Node.js, where it is not dynamically downloaded, for example in `js-revive`

.github/workflows/build-revive-wasm.yml
smiasojed: Does not make sense. I am not sure if I will merge it, maybe I should find better solution

smiasojed: Renamed to resolc_web.js as it is going to be used only from browser

PR #180, opened 2025-01-29
Align emscripten compilation options

smiasojed:


.cargo/config.toml
```diff
@@ -12,8 +12,6 @@ rustflags = [
     "-Clink-arg=-sALLOW_TABLE_GROWTH=1",
     "-Clink-arg=--js-library=js/embed/soljson_interface.js",
     "-Clink-arg=--pre-js=js/embed/pre.js",
-    "-Clink-arg=-sNODEJS_CATCH_EXIT=0",
-    "-Clink-arg=-sDISABLE_EXCEPTION_CATCHING=0",
     "-Clink-arg=-sSTACK_SIZE=128kb",
-    "-Copt-level=3"
+    "-Clink-arg=-sNODEJS_CATCH_EXIT=0"
```
xermicus, line 16: Are you removing the opt level on purpose?

smiasojed: I see in logs that compiling in release mode adds this option automatically

xermicus: Thanks for the clarification.

xermicus (approved):

crates/llvm-builder/src/platforms/wasm32_emscripten.rs
```diff
@@ -142,6 +142,7 @@ fn build_target(
             .env("EMCC_DEBUG", "2")
             .env("CXXFLAGS", "-Dwait4=__syscall_wait4")
             .env("LDFLAGS", "-lnodefs.js -s NO_INVOKE_RUN -s EXIT_RUNTIME -s INITIAL_MEMORY=64MB -s ALLOW_MEMORY_GROWTH -s EXPORTED_RUNTIME_METHODS=FS,callMain,NODEFS -s MODULARIZE -s EXPORT_ES6 -s WASM_BIGINT")
```
xermicus, line 144: This should be removed?

smiasojed: yes, it should. Thanks!

Issue #177, opened 2025-01-28
Revive JS version fails when compiling contract which instantiate another one


smiasojed:
Contract:
```
// SPDX-License-Identifier: GPL-3.0

pragma solidity >=0.8.2 <0.9.0;

contract C1 {
    constructor() {
    }
}

contract C2 {
    constructor() {
        C1 newContract = new C1();
    }
}
```

Callstack after compiling to asm.js target:
```
     Error: Aborted(segmentation fault)
      at abort (examples/node/resolc.js:8848881:41)
      at segfault (examples/node/resolc.js:8848579:3)
      at $75436 (examples/node/resolc.js:8847440:4)
      at emscripten_builtin_malloc (examples/node/resolc.js:8796570:6)
      at llvm__SmallVectorBase_unsigned_20int___grow_pod_28void__2c_20unsigned_20long_2c_20unsigned_20long_29 (examples/node/resolc.js:1608844:13)
      at $28anonymous_20namespace_29__DAGCombiner__visitTokenFactor_28llvm__SDNode__29 (examples/node/resolc.js:4496984:17)
      at $28anonymous_20namespace_29__DAGCombiner__combine_28llvm__SDNode__29 (examples/node/resolc.js:4421389:10)
      at llvm__SelectionDAG__Combine_28llvm__CombineLevel_2c_20llvm__AAResults__2c_20llvm__CodeGenOptLevel_29 (examples/node/resolc.js:4420692:16)
      at llvm__SelectionDAGISel__CodeGenAndEmitDAG_28_29 (examples/node/resolc.js:4962737:6)
      at llvm__SelectionDAGISel__SelectBasicBlock_28llvm__ilist_iterator_w_bits_llvm__ilist_detail__node_options_llvm__Instruction_2c_20true_2c_20false_2c_20void_2c_20true__2c_20false_2c_20true__2c_20llvm__ilist_iterator_w_bits_llvm__ilist_detail__node_options_llvm__Instruction_2c_20true_2c_20false_2c_20void_2c_20true__2c_20false_2c_20true__2c_20bool__29 (examples/node/resolc.js:4964258:6)
      at llvm__SelectionDAGISel__SelectAllBasicBlocks_28llvm__Function_20const__29 (examples/node/resolc.js:4961395:22)
      at llvm__SelectionDAGISel__runOnMachineFunction_28llvm__MachineFunction__29 (examples/node/resolc.js:4958283:35)
[...]

PR #184, opened 2025-01-30
Fix stack overflow issue

smiasojed:
Fixes: https://github.com/paritytech/revive/issues/177
Default stack size is 64KB which is not enough for our use cases


xermicus (commented):
Nice catch!

You got any insights on the stack size? We just might give it some more margin. Can we add a second fixture with multiple different contract instantiations?

smiasojed:
> Nice catch!
> 
> You got any insights on the stack size? We just might give it some more margin. Can we add a second fixture with multiple different contract instantiations?

I enable heap and stack checks and allocations logging to see the issue

xermicus:
> what sort of test would you like to have

Can you add both? :slightly_smiling_face:

smiasojed:
Finally I decided to add a test:
```
  constructor() {
    new C1();
    new C2();
    new C3();
    new C4();
    new C5();
  }
```
they are both very similar in this context

xermicus (approved):

Issue #66, opened 2024-10-04
Backport bug fixes from the era compiler


xermicus:
The API is already available in the pallet and should work fine. Once this is implemented we should test it with the openzeppelin proxy implementations.

PR #186, opened 2025-01-30
remove support for legacy evm assembly 

### LINKED ISSUE #89: Fix or remove EVMLA
From the [ZKSync docs](https://docs.zksync.io/zk-stack/components/compiler/specification/evmla-translator):

> EVM legacy assembly is very challenging to translate to LLVM IR, since it obfuscates the control flow of the program and uses a lot of dynamic jumps. Most of the jumps can be translated to static ones by using a static analysis of EVM assembly, but some of jumps are impossible to resolve statically. For example, internal function pointers can be written to memory or storage, and then loaded and called. Recursion is another case we have skipped for now, as there is another stack frame allocated on every iteration, preventing the static analyzer from resolving the jumps.
> 
> Both issues are being worked on in our fork of the Solidity compiler, where we are changing the codegen to remove the dynamic jumps and add the necessary metadata.

Their solc fork is in the works for a long time now (months if not years) and this is still the status quo. Also ZKSync is an ETH L2 and since they got away with this so far.

I'd like to set a clear path forward.
- It is not entirely clear what the use case is for this? YUL is available and considered stable enough for years.
- If we ultimately need it we can either
  - Declare it as experimental and accept the current status quo
  - Properly fix it on our own
  - Possibly work around this by just resolving those jumps dynamically at runtime? 
  - Direct people to the matter-labs solc fork, assuming it is still general enough

xermicus:
The initial `resolc` release supports YUL only. This PR removes dead code. For more details see #89. In a nutshell,  EVMLA translation is:
- Incomplete
- Obsolete
- Emitting inefficient code

As a consequence, `solc` versions without or buggy YUL support (< `0.8`) are no longer supported.

Closes #89
Partially addresses #66


athei (approved):
Yes. For now we don't support. If we really need it we will add it layer.

PR #207, opened 2025-02-10
LLVM release workflows

mutantcornholio:
I've been testing these on a fork with public runners.  
Windows, unfortunately, still needs work, and linux fails on a 6h timeout, when executed on ubuntu-latest 😅.  
But as at least some of it works, let's kick off the review, and hopefully, I'll be able to finish testing workflows and windows LLVM release in parallel (given the duration of the feedback loop on these tasks)

### DIFF
--- a/.github/workflows/release-llvm.yml
+++ b/.github/workflows/release-llvm.yml
@@ -1,0 +1,155 @@
+name: Release LLVM
+
+on:
+  push:
+    tags:
+      - "llvm*"
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: create release
+        uses: softprops/action-gh-release@v2
+        with:
+          name: "LLVM binaries release: ${{ github.ref_name }}"
+          body: "This release includes binaries of LLVM, used to compile revive itself"
+          make_latest: "false"
+
+  build-macos:
+    strategy:
+      matrix:
+        os: [macos-14, macos-13]
+        include:
+          - os: macos-13
+            arch: x64
+          - os: macos-14
+            arch: arm64
+    needs: create-release
+    runs-on: ${{ matrix.os }}
+    name: "build-macos-${{ matrix.arch }}"
+    env:
+      RUST_LOG: trace
+    permissions:
+      contents: write # for uploading assets to release
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: install macos deps
+        run: |
+          brew install ninja
+
+      - name: versions
+        run: |
+          rustup show
+          cargo --version
+          cmake --version
+          echo "bash:" && bash --version
+          echo "ninja:" && ninja --version
+          echo "clang:" && clang --version
+
+      - name: Build LLVM
+        run: |
+          make install-llvm
+
+      - name: clean
+        # check removed files
+        run: |
+          cd target-llvm/gnu/target-final/bin/
+          rm diagtool llvm-libtool-darwin llvm-lipo llvm-pdbutil llvm-dwarfdump llvm-nm llvm-readobj llvm-cfi-verify \
+             sancov llvm-debuginfo-analyzer llvm-objdump llvm-profgen llvm-extract llvm-jitlink llvm-c-test llvm-gsymutil llvm-dwp \
+             dsymutil llvm-dwarfutil llvm-exegesis lli clang-rename bugpoint clang-extdef-mapping clang-refactor c-index-test \
+             llvm-reduce llvm-lto clang-linker-wrapper llc llvm-lto2
+
+      - name: package artifacts
+        run: |
+          tar -czf "${{ github.ref_name }}-macos-${{ matrix.arch }}.tar.gz" target-llvm/gnu/target-final
+
+      - name: upload archive to release
+        uses: softprops/action-gh-release@v2
+        with:
+          make_latest: "false"
+          files: |
+            ${{ github.ref_name }}-macos-${{ matrix.arch }}.tar.gz
+
+
+  build-linux-all:
+    needs: create-release
+    runs-on: parity-large
+    env:
+      RUST_LOG: trace
+    permissions:
+      contents: write # for uploading assets to release
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: install linux deps
+        run: |
+          sudo apt-get update && sudo apt-get install -y cmake ninja-build curl git libssl-dev pkg-config clang lld musl
+
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: stable
+          components: rust-src
+          target: wasm32-unknown-emscripten
+          rustflags: ""
+
+      - name: versions
+        run: |
+          rustup show
+          cargo --version
+          cmake --version
+          echo "bash:" && bash --version
+          echo "ninja:" && ninja --version
+          echo "clang:" && clang --version
+
+      - name: Build host LLVM
+        run: |
+          make install-llvm
+
+      - name: Build gnu LLVM
+        run: |
+          revive-llvm clone
+          revive-llvm build --llvm-projects lld --llvm-projects clang
+
+      - name: Build musl LLVM
+        run: |
+          revive-llvm --target-env musl build --llvm-projects lld --llvm-projects clang
+
+      - name: Build emscripten LLVM
+        run: |
+          revive-llvm --target-env emscripten clone
+          source emsdk/emsdk_env.sh
+          revive-llvm --target-env emscripten build --llvm-projects lld
+
+      - name: clean
+        # check removed files
+        run: |
+          for target in gnu emscripten musl; do
+            cd target-llvm/${target}/target-final/bin/
+            rm -rf diagtool llvm-libtool-darwin llvm-lipo llvm-pdbutil llvm-dwarfdump llvm-nm llvm-readobj llvm-cfi-verify \
+               sancov llvm-debuginfo-analyzer llvm-objdump llvm-profgen llvm-extract llvm-jitlink llvm-c-test llvm-gsymutil llvm-dwp \
+               dsymutil llvm-dwarfutil llvm-exegesis lli clang-rename bugpoint clang-extdef-mapping clang-refactor c-index-test \
+               llvm-reduce llvm-lto clang-linker-wrapper llc llvm-lto2
+            cd -
+          done
+
+      - name: package artifacts
+        run: |
+          tar -czf "${{ github.ref_name }}-x86_64-linux-gnu-linux.tar.gz" target-llvm/gnu/target-final
+          tar -czf "${{ github.ref_name }}-x86_64-linux-musl.tar.gz" target-llvm/musl/target-final
+          tar -czf "${{ github.ref_name }}-wasm32-unknown-emscripten.tar.gz" target-llvm/emscripten/target-final
+
+      - name: upload archive to release
+        uses: softprops/action-gh-release@v2
+        with:
+          make_latest: "false"
+          files: |
+            ${{ github.ref_name }}-x86_64-linux-gnu-linux.tar.gz
+            ${{ github.ref_name }}-x86_64-linux-musl.tar.gz
+            ${{ github.ref_name }}-wasm32-unknown-emscripten.tar.gz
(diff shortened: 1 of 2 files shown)


xermicus:
In #201 it was discussed to not do LLVM releases, now I'm a bit confused 😅 Do you plan to have separate workflows, but the binary release workflow is depending on this first? That would make sense to me (i.e. a release tag is first causing the LLVM workflows to run, and afterwards the binary is build using those artifacts - but the LLVM artifacts and the binary artifacts end up in the same release).

> linux fails on a 6h timeout, when executed on ubuntu-latest 

Yeah those are absolute potato VMs, I suggest using our parity runners.

xermicus:
> I don't see why we should release LLVM on each release of revive.

What you are suggesting is that we use two tags to trigger either the LLVM build or the `resolc` binaries release build, right? And the `resolc` release workflow always downloads the latest (or specified) LLVM build for each platform, instead of compiling it on it's own.

LLVM is just a build artifact. In that sense we don't release LLVM. We depend on it and attach it to the release as artifacts for convenience reasons. Having a dedicated tag and release for LLVM vs. releasing the artifacts together with the binary release seems mostly bikeshedding to me, with dedicated release increasing complexity. 

> Even with cache, it seems a kinda wasteful.

I'm unsure about the added complexity is worth it (still unclear to me what exactly we gain from dedicated LLVM releases), but if you think of it as the better solution go for it. Keep in mind that every `polkadot-sdk` PR likely uses more compute than a full release build for all platforms.

athei:
> LLVM is just a build artifact. In that sense we don't release LLVM. We depend on it and attach it to the release as artifacts for convenience reasons. Having a dedicated tag and release for LLVM vs. releasing the artifacts together with the binary release seems mostly bikeshedding to me, with dedicated release increasing complexity.

If we don't create a release where would we store the LLVM artifact so that it can be downloaded by the revive release?

athei:
> We just cache it as we do currently.

Ahh okay you mean basically no new workflow and we just have https://github.com/paritytech/revive/pull/201 ?

xermicus:
Yeah this is what I meant. But having it separate is also fine for me.

athei:
So tl;dr we are fine with this PR and just waiting for macOS and windows to get fixed?

xermicus:
Yes I'm fine, waiting for everything fixed and `main` merged and then I'll give it a proper review.

mutantcornholio:
MacOS does work already, but Windows builds are inconsistent. Trying to figure them out.

mutantcornholio:
Should we maybe go forward and do the first iteration without windows? The issue with windows is not trivial, and the feedback loop on windows runners is two hours.  
If we can have llvm release pipeline working for linux and mac, we can go forward with integrating it with testing pipeline and releasing pipeline for linux and mac, which would bring the end result closer.

.github/workflows/release-llvm.yml
```diff
@@ -0,0 +1,205 @@
...
+#          install: >-
+#            base-devel
+#            mingw-w64-x86_64-clang
+#            mingw-w64-x86_64-lld
+#            mingw-w64-x86_64-rust
+#            mingw-w64-x86_64-cmake
+#            mingw-w64-x86_64-gcc
+#            mingw-w64-x86_64-gcc-libs
+#            mingw-w64-x86_64-python
+#            mingw-w64-clang-x86_64-riscv64-unknown-elf-toolchain
```
athei, line 165: Use msvc since you are building on windows. It is the most compatible and easiest to build for natively. No reason to pull mingw.

athei (changes requested):

athei (changes requested):
- Windows should use MSVC
- macOS needs to be a universal binary: Build twice: Once on a Intel runner and once on a AppleSilicon runner. Then use `lipo` to merge the two binaries. You can also build both on any macOS runner since cross compiling in macOS is easy.

athei:
> Should we maybe go forward and do the first iteration without windows? 

To me it does not make sense to ship this incomplete. We don't gain anything from that because we don't even have the revive workflow done. So there are no users of this.

> The issue with windows is not trivial, and the feedback loop on windows runners is two hours.

Switching to the default toolchain on windows (msvc) hopefully makes it easier.

.github/workflows/release-llvm.yml
mutantcornholio: As a matter of fact, I've started with msvc and failed. The suggestion from @xermicus was to look for inspiration in how matter-labs do their windows releases, and [they use mingw](https://github.com/matter-labs/era-compiler-ci/blob/main/.github/actions/prepare-msys/action.yml)

mutantcornholio: I think [this](https://github.com/llvm/llvm-project/issues/47189) was the issue that I've been hitting with MSVC

mutantcornholio: > I would be really surprised if nobody could build LLVM with the Microsoft toolchain.

I wouldn't be surprised if nobody has built matterlabs fork of LLVM with MS toolchain, as they're using mingw themselves.  
Given the sort of issue I'm getting right now, however, I'm inclined to give it another go.

mutantcornholio:
> > I don't get it. These artifacts will be used in the revive workflow. Revive workflow depends on it, not the other way around. Why would revive workflow be a blocker here?
> 
> So you would change the in-progress other PR then?

Not sure if I understand what you mean, but workflows in other PRs would be changed to use artifacts produced by this one. Is that what you're asking?

athei:
Yes. I am just saying me mindful of that potential interaction with that other open PR.

mutantcornholio:
That interaction is intended and planned from the start

xermicus:
LLVM is only a dependency. So yes, if you cross-compile you need to cross-compile everything.

athei:
> > * macOS needs to be a universal binary: Build twice: Once on a Intel runner and once on a AppleSilicon runner. Then use `lipo` to merge the two binaries. You can also build both on any macOS runner since cross compiling in macOS is easy.
> 
> You mean resolc, not llvm though, right? If cross-compiling is easy, we wouldn't need to create a universal binary of every llvm / clang tool, right?

LLVM is an intermediate artifact statically linked by the `resolc` binary. So it is not really important for them to be universal as they are not used directly by users. If that is easier (it probably is) you can just can do two separate artifacts for LLVM in this workflow: One for x64 and one for aarch64. But the point is that have to build it twice for macOS in this workflow. Either on different runners or via cross compilation on the same runner.

You **could** unify those two LLVM artifacts then by making all the binaries and libs universal. But you are right: I wouldn't do that. It's not worth it. We'll only do that for `resolc`.

athei:
Let's skip the windows build for now and do that as a follow up. macOS and Linux is enough for now since those seem to be not problematic. I will look into the Windows build when the other things are working.

mutantcornholio:
I'll take this to draft for the time being, after #201 is merged, I'll alter that release pipeline to use llvm from this one, and hopefully add macos to it as well.

mutantcornholio:
Okay, here's an example of how LLVM release looks (except for the `latest` label, that shouldn't happen anymore): https://github.com/paritytech/revive-workflow-test/releases/tag/llvm-0.0.4.  
I've [tested](https://github.com/paritytech/revive-workflow-test/actions/runs/13368591564?pr=6) building revive with it on linux (musl, emscripten) and macos (intel, arm).  
Currently working on making a universal binary for mac, and other workflow fixes, but this PR is ready for review (and we'll need to have first LLVM release done before we can use it in the main pipeline).

crates/llvm-builder/src/builtins.rs
athei, line 1: Please remove changes here as we are doing nothing windows related in this PR.

crates/llvm-builder/src/lib.rs
```diff
@@ -302,7 +302,6 @@ pub fn build(
     } else {
         anyhow::bail!("Unsupported target architecture");
     }
-
```
athei, line 305: Unrelated change please roll back.

.github/workflows/release-llvm.yml
```diff
@@ -0,0 +1,231 @@
...
+
+  build-macos:
+    strategy:
+      matrix:
+        os: [macos-14, macos-13]
+        include:
+          - os: macos-13
+            arch: x64
+          - os: macos-14
+            arch: arm64
```
athei, line 30: Why are we using different macOS versions for the different archs? The latest version does support Intel.

.github/workflows/release-llvm.yml
```diff
@@ -0,0 +1,231 @@
...
+#      - name: package artifacts
+#        run: |
+#          tar -cf "${{ github.ref_name }}-windows-gnu.tar.gz" target-llvm/gnu/target-final
+#
+#      - name: upload archive to release
+#        uses: softprops/action-gh-release@v2
+#        with:
+#          make_latest: false
+#          files: |
+#            ${{ github.ref_name }}-windows-gnu.tar.gz
```
athei, line 231: Please don't commit commented out code.

athei:
> The linux release you linked contains musl and emscripten; I think it would be much clearer if there were three separate artifacts (emscirpten, linux-musl and linux-gnu)?

Makes sense. Except that we are not building for gnu. Only musl.

.github/workflows/release-llvm.yml
mutantcornholio: Because that's the only available option as far as github runners go: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

mutantcornholio:
> The linux release you linked contains musl and emscripten; I think it would be much clearer if there were three separate artifacts (emscirpten, linux-musl and linux-gnu)?

We aren't going to compile musl and emscripten in separate jobs. Separating artifacts would just complicate things.

.github/workflows/release-llvm.yml
athei: thanks make sense

crates/llvm-builder/src/builtins.rs
mutantcornholio: done

xermicus:
Sorry maybe that wasn't too clear. MUSL is what we build release builds against. For development, cross-compiling all the time is way too cumbersome. The CI could be set up to always cross-compile but I'm not sure if it is worth it, so it will rely on the GNU build. Additionally, the GNU build can be used conveniently by people not wanting to build LLVM locally (slow laptop or whatever reason).

mutantcornholio:
So we do expect people to download LLVM binaries to use it for development?  
@athei what's your take on this? 

I mean, if we do, then we probably should also provide emscripten version for mac as well.

mutantcornholio:
> I don't understand, why emscripten cross compiled on macos?

Is it like, not possilble, or devs wouldn't do that on their machines?

mutantcornholio:
Ah, so devs would download LLVM for their specific platform (linux / mac intel / mac arm) plus emscripten build separately.  
Okay, now I see the reason to have emscripten as a separate artifact.

xermicus:
Yeah that's the idea. People (or CI jobs) who don't feel like compiling their own LLVM builds need to get the LLVM build for whatever platform they need :) So please just add the GNU LLVM build too - so that it will be available for the linux CI runners and Linux folks. While we might switch the CI, never expect people on Linux to cross-compile this, so we need it anyways.

mutantcornholio:
It seems that there's some issue on GitHub API side on creating non-latest releases: https://github.com/orgs/community/discussions/151811

In order not to mark LLVM releases latest for now, how about we make the action create them as draft, and we'll publish them manually afterwards? Manual publish without marking release as latest works.

athei:
> how about we make the action create them as draft, and we'll publish them manually afterwards?

Yes this is a sensible thing to do anyways.

athei:
Looks good now. Can you do the tag on your fork so we can check the generated artifacts?

mutantcornholio:
> Looks good. Can you do the tag on your fork so we can check the generated artifacts?

https://github.com/paritytech/revive-workflow-test/releases/tag/llvm-0.0.9

AndWeHaveAPlan (approved):

mutantcornholio:
@AndWeHaveAPlan noticed that I forgot to add compression flag to archives; PR's updated, testing it out in [this workflow](https://github.com/paritytech/revive-workflow-test/actions/runs/13420960900)

.github/workflows/release-llvm.yml
```diff
@@ -0,0 +1,154 @@
...
+        uses: softprops/action-gh-release@v2
+        with:
+          files: |
+            ${{ github.ref_name }}-macos-${{ matrix.arch }}.tar.gz
+
+
+  build-linux-all:
+    needs: create-release
+    runs-on: parity-large
+    #    runs-on: ubuntu-latest
```
xermicus, line 83:
```suggestion
```

athei (approved):

PR #215, opened 2025-02-17
ci: do not run bun tests

xermicus:
They are too flaky. Culprit: https://github.com/oven-sh/bun/issues/17146

Updates some JS dependencies as a drive-by.


smiasojed (approved):

athei:
Shouldn't we just use a non broken runtime, instead? Would it be hard to port to `node`?

xermicus:
@athei FYI the PR pivoted already to do exactly that (originally I thought it was fixed with upgraded bun because the error went away locally after upgrading bun but then I got remember of the underlying issue).

.github/workflows/build-revive-wasm.yml
```diff
@@ -118,16 +117,9 @@ jobs:
         run: |
           Set-ExecutionPolicy RemoteSigned -Scope CurrentUser
           iex (new-object net.webclient).downloadstring('https://get.scoop.sh')
-          scoop install bun@${{ env.BUN_VERSION }}
```
athei, line 121: Name of the job still correct?

athei (commented):
So we are running the tests using node anyways? We just removed the additional run via bun?

.github/workflows/build-revive-wasm.yml
xermicus: Good catch thanks! I probably should have removed the whole thing instead.

smiasojed: I think that you need to keep wget

xermicus: Ah so it wasn't used only for downloading bun?

PR #220, opened 2025-02-19
Using released LLVM in revive build & MacOS revive pipeline & pipeline improvements

mutantcornholio:
This PR depends on #207, so we'll need to merge that one first

Here's how the workflow and release look in fork:
https://github.com/paritytech/revive-workflow-test/actions/runs/13442528853
https://github.com/paritytech/revive-workflow-test/releases/tag/untagged-c2d093272cba97f2eb1c

List of changes: 
* Instead of compiling LLVM, it downloads it from releases, using `.github/workflows/get-release-artifact.js` in `get-llvm-artifact` step, with `curl` / `tar` following
* Added MacOS build with generating unified binary at the end
* Moved release notes generation to the first step to avoid `checkout` at the end
* Joined `create-release` and `upload-assets` jobs into a single `create-release` job
* Replaced archived `actions/upload-release-asset` action and custom release creation with a single call of `softprops/action-gh-release@v2`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,7 @@ run-name: Release ${{ github.ref_name }}
 on:
   push:
     branches:
-      - 'main'
+      - "main"
 
 concurrency:
   group: ${{ github.ref }}-${{ github.workflow }}
@@ -12,12 +12,9 @@ concurrency:
 env:
   #rust-musl-cross:x86_64-musl
   RUST_MUSL_CROSS_IMAGE: messense/rust-musl-cross@sha256:68b86bc7cb2867259e6b233415a665ff4469c28b57763e78c3bfea1c68091561
+  RUST_LOG: trace
 
 jobs:
-
-  #
-  #
-  #
   tag:
     runs-on: ubuntu-24.04
     permissions:
@@ -25,11 +22,12 @@ jobs:
     outputs:
       TAG: ${{ steps.versions.outputs.TAG }}
       PKG_VER: ${{ steps.versions.outputs.PKG_VER }}
+      RELEASE_NOTES: ${{ steps.versions.outputs.RELEASE_NOTES }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          fetch-tags: 'true'
+          fetch-tags: "true"
           fetch-depth: 0
 
       - name: Versions
@@ -41,7 +39,7 @@ jobs:
           echo "Package version $PKG_VER"
           #
           echo "PKG_VER=$PKG_VER" >> $GITHUB_OUTPUT
-          if [ $CURRENT_TAG == $PKG_VER  ];
+          if [[ $CURRENT_TAG == $PKG_VER  ]];
             then
               echo "Tag is up to date. Nothing to do.";
               export TAG=old;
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -1,347 +1,356 @@
...
+          tag_name: ${{ needs.tag.outputs.PKG_VER }}
+          name: ${{ needs.tag.outputs.PKG_VER }}
+          draft: true
+          prerelease: true # doublecheck if we need prerelease here
+          files: |
+            ./resolc-linux/resolc-static-linux
+            ./resolc-macos/resolc-macos
+            ./resolc-wasm/resolc.js
+            ./resolc-wasm/resolc.wasm
+            ./resolc-wasm/resolc_web.js
```
athei, line 356: I know it is shaving the yak here but I didnt realize earlier. Lets pack up those artifacts:
- Linux, macOS, Windows resolc should be compressed wit xz
- The three emscripten artifacts should be tar.xz into one archive.

mutantcornholio: > Linux, macOS, Windows resolc should be compressed wit xz

I see how 50+Mb binaries are better to be compressed, but I'd probably go with `gzip` instead of `xz`, as it's not a builtin tool in macos.  

For windows (when we'll do it) I'd even go with standard `.zip` archive.

mutantcornholio:
Here's how the release looks after updates
https://github.com/paritytech/revive-workflow-test/actions/runs/13442528853
https://github.com/paritytech/revive-workflow-test/releases/tag/untagged-c2d093272cba97f2eb1c

.github/workflows/release.yml
```diff
@@ -1,347 +1,367 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-tags: "true"
+          fetch-depth: 0
+
+      - name: Versions
+        id: versions
+        run: |
+          export CURRENT_TAG=$(git describe --tags --abbrev=0 --exclude "llvm-*")
```
AndWeHaveAPlan, line 36:
```suggestion
          export CURRENT_TAG=$(git describe --tags --abbrev=0 --exclude "llvm-*" --exclude "*-revive*")
```

there is no `llvm-` prefix for llvm releases now, so `release` will be triggered on every push after llvm tag
https://github.com/paritytech/revive/actions/runs/13454276537/job/37594743538#step:3:2

xermicus: I think the proper fix is to make sure we retain the llvm tag: https://github.com/paritytech/revive/pull/224

mutantcornholio: Ah god dammit, the tag should've been `llvm-18.1.8-revive.6ad7908`...

.github/workflows/release.yml
```diff
@@ -1,347 +1,367 @@
...
+          path: resolc-macos/
+
+      - name: create-release
+        uses: softprops/action-gh-release@v2
+        with:
+          body: ${{ needs.tag.outputs.RELEASE_NOTES }}
+          tag_name: ${{ needs.tag.outputs.PKG_VER }}
+          name: ${{ needs.tag.outputs.PKG_VER }}
+          draft: true
+          prerelease: true # doublecheck if we need prerelease here
```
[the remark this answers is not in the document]
xermicus: Yes let's only do draft
```suggestion
```

xermicus (approved):
LGTM

PR #225, opened 2025-02-21
Fix: add llvm- prefix to llvm version tag

mutantcornholio:
Needed both for differentiating them from regular releases and for
filtering the tags in automation


AndWeHaveAPlan:
This one will also trigger the `release` workflow. Whoever merge this please also cancel release

PR #228, opened 2025-02-24
Add per file output selection for `--standard-json` mode and some other small foundry support bits.

pkhry:
### Description

This PR adds per file support for output selection in `--standard-json` inputs this is needed for foundry support. 

This PR also adds `Ir` and `Unknown` flags to the output selection options, ir for the unoptimized IR and `Unknown` to catch unsupported options as they would not be applied anyways.(Although on take 2 it seems like it's better to error out explicitely rather than drop unsupported flags.)

### DIFF
--- a/crates/solidity/src/solc/standard_json/input/settings/selection/file/flag.rs
+++ b/crates/solidity/src/solc/standard_json/input/settings/selection/file/flag.rs
@@ -43,9 +43,6 @@ pub enum Flag {
     /// The Ir
     #[serde(rename = "ir")]
     Ir,
-    // Catch all unsupported flags
-    #[serde(other)]
-    Unknown,
 }
 
 impl std::fmt::Display for Flag {
@@ -64,7 +61,6 @@ impl std::fmt::Display for Flag {
             Self::EVMDBC => write!(f, "evm.deployedBytecode"),
             Self::Assembly => write!(f, "evm.assembly"),
             Self::Ir => write!(f, "ir"),
-            Self::Unknown => write!(f, "unknown flag"),
         }
     }
 }
--- a/crates/solidity/src/solc/standard_json/input/settings/selection/file/flag.rs
+++ b/crates/solidity/src/solc/standard_json/input/settings/selection/file/flag.rs
@@ -43,7 +43,7 @@ pub enum Flag {
     /// The Ir
     #[serde(rename = "ir")]
     Ir,
-    // Catch all unsupported functions
+    // Catch all unsupported flags
     #[serde(other)]
     Unknown,
 }
--- a/crates/solidity/src/solc/standard_json/input/settings/selection/file/flag.rs
+++ b/crates/solidity/src/solc/standard_json/input/settings/selection/file/flag.rs
@@ -40,6 +40,12 @@ pub enum Flag {
     /// The assembly code
     #[serde(rename = "evm.assembly")]
     Assembly,
+    /// The Ir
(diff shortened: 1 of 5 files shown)


crates/solidity/src/solc/standard_json/input/settings/selection/file/flag.rs
```diff
@@ -40,6 +40,12 @@ pub enum Flag {
     /// The assembly code
     #[serde(rename = "evm.assembly")]
     Assembly,
+    /// The Ir
+    #[serde(rename = "ir")]
+    Ir,
+    // Catch all unsupported flags
+    #[serde(other)]
+    Unknown,
```
xermicus, line 48: I don't like this. I'd prefer to be explicit about the output selection. If there is a bug we can fix it and if `solc` adds new one we should catch it and implement those properly.

xermicus (changes requested):

crates/solidity/src/solc/standard_json/input/settings/selection/file/flag.rs
pkhry: removed it

xermicus (approved):
LGTM, thank you

xermicus:
> is needed for foundry support

What does foundry do with it? I don't see why tooling needs this.

crates/solidity/src/solc/standard_json/input/settings/selection/file/flag.rs
pkhry: my bad, forgot that serde already generates tolerable error messages for simple enums.

crates/solidity/src/solc/standard_json/input/settings/selection/file/flag.rs
```diff
@@ -57,6 +60,7 @@ impl std::fmt::Display for Flag {
             Self::EVMBC => write!(f, "evm.bytecode"),
             Self::EVMDBC => write!(f, "evm.deployedBytecode"),
             Self::Assembly => write!(f, "evm.assembly"),
+            Self::Ir => write!(f, "ir"),
```
smiasojed, line 63: we have evm.bytecode, so why it is not evm.ir?

xermicus: `solc` puts EVM related artifacts into the `evm` objects ([see the docs for where is what](https://docs.soliditylang.org/en/latest/using-the-compiler.html#output-description)).

So the IR is probably not there because it isn't directly tied to EVM. Even` though it could because supposedly there are multiple YUL dialects, but this never manifested and there is just the EVM YUL dialect. If there would be a second one it's unclear to me where they would put it because the root object already implicitly contains the YUL _EVM_ IR dialect. But we don't need to worry about it since if it ever happens we just adapt to whatever they are doing.

pkhry: ir is just an unoptimized YUL output that might be used down the line in `foundry` itself and it seemed like a harmless thing to add support for.

xermicus: `revive` does process the optimized IR. So why does foundry need it?

pkhry: `Ir` is used for forge inspect and that's basically it.

PR #230, opened 2025-02-24
CI improvements

AndWeHaveAPlan:
- llvm artifacts search, download and extract now in `get-llvm` action 
- `get-emsdk` action - clone, install, activate emsdk, so we don't need to build `revive-builder` for this
- switch workflows to new actions
- `concurrency` for remaining workflows (cancel run if new run is triggered)
- `target_commitish` for main release, fixes release commit 
- Run release workflow in PR without creating a release if PR is labeled with `release-test`


AndWeHaveAPlan:
[fork run](https://github.com/paritytech/revive-workflow-test/actions/runs/13495620252)
[fork published release](https://github.com/paritytech/revive-workflow-test/releases/tag/v0.1.0-dev.11)

.github/actions/get-llvm/action.yml
```diff
@@ -0,0 +1,91 @@
...
+      env:
+        releasePrefix: ${{ inputs.releasePrefix }}
+        artifactArch: ${{ inputs.artifactArch }}
+      with:
+        result-encoding: string
+        script: |
+          let page = 1;
+          let releases = [];
+
+          let releasePrefix = process.env.releasePrefix
```
athei, line 43: Since the script is now embedded shouldn't we delete the `get-release-artifact.js` file?

AndWeHaveAPlan: Yes, it's no longer needed, deleted

athei (commented):
Screenshot 2025-02-24 at 17 43 19


Can't mark the line in the script. But this isn't correct, right? This should be expanded to the actual arch? In the script it uses `"` which might cause this?

Another question: The `release-test` lable only does something in case you actually also bump the version side the PR, right?

AndWeHaveAPlan:
> Can't mark the line in the script. But this isn't correct, right? This should be expanded to the actual arch? In the script it uses `"` which might cause this?

No, its because the job is skipped, so gh does not resolve `matrix`
Just removed arch part, as gh automatically expand this on job run, like for `test-revive-wasm (*)`


> Another question: The `release-test` lable only does something in case you actually also bump the version side the PR, right?

Yes, for PR's workflow will run like a real release, except for the creation part

athei (approved):
Thanks. Let's have @xermicus have a look before merging.

.github/workflows/release.yml
```diff
@@ -203,10 +180,12 @@ jobs:
           echo "ninja:" && ninja --version
           echo "clang:" && clang --version
 
-      - name: download llvm
-        run: |
-          curl -L -o llvm.tar.gz "${{ steps.get-llvm-musl-artifact.outputs.result }}"
-          tar -xvf llvm.tar.gz
+      - name: get llvm gnu
```
xermicus, line 183:
```suggestion
      - name: get llvm musl
```

I'm seeing musl below

.github/workflows/release.yml
```diff
@@ -247,29 +226,16 @@ jobs:
         with:
           node-version: "20"
 
-      - name: Get latest emscripten LLVM release artifact
-        id: get-llvm-emscripten-artifact
-        uses: actions/github-script@v7
+      - name: get llvm gnu
```
xermicus, line 229:
```suggestion
      - name: get llvm gnu
```

I see emscripten below

xermicus (approved):
LGTM

PR #234, opened 2025-02-25
llvm-context: modularize compiler builtin functions

xermicus:
- Add the revive runtime function interface to minimize boiler plate code.
- Outline heavily repeated code into dedicated functions to bring down code size.
- The code size tests builds optimized for size.
- Function attributes are passed as slices.

This significantly brings down the code size for all OpenZeppelin wizard contracts (using all possible features) compiled against OpenZeppelin `v5.0.0` with size optimizations.

|contract|| `-Oz` main | `-Oz` PR || `-O3` main | `-O3` PR |
|-|-|-|-|-|-|-|
|erc1155.sol||100K|67K||114K|147K|
|erc20.sol||120K|90K||160K|191K|
|erc721.sol||128K|101K||178K|214K|
|governor.sol||226K|165K||293K|349K|
|rwa.sol||116K|85K||154K|185K|
|stable.sol||116K|86K||155K|192K|

On the flip side this introduces a heavy penalty for cycle optimized builds. Setting the no-inline attributes for cycle optimized builds helps a lot but heavily penalizes runtime speed (LLVM does not yet inline everything properly - to be investigated later on). 

Next steps:
- Modularize more functions
- Refactor the YUL function arguments to use pointers instead of values
- Afterwards check if LLVM still has trouble inline-ing properly on O3 or set the no-inline attribute if it does not penalize runtime performance too bad.


athei:
Noice.

> Refactor the YUL function arguments to use pointers instead of values

Yeah passing account ids around was also making ink! contracts really big. You would think llvm could demote those to reference by itself given that it knows that they don't alias in Rust but it doesn't.

xermicus:
> Yeah passing account ids around was also making ink! contracts really big. You would think llvm could demote those to reference by itself given that it knows that they don't alias in Rust but it doesn't.

Yeah I agree, the devil is probably in the details.

PR #240, opened 2025-02-26
Fixing revive release artifacts

mutantcornholio:
* Stripping "./resolc-out" from linux artifact is weirdly
`--strip-components 2`, not `1`
* Adding execution bit for macos binary


mutantcornholio:
closing in favor of #241

PR #250, opened 2025-02-28
Build resolc for windows

athei:


athei:
Meant to open against my fork. Will sed new PR when thoroughly tested.

Issue #162, opened 2025-01-15
Expose the supported compiler versions on GitHub Pages


smiasojed:
Expose the compiler versions for all supported platforms on GH pages. You can take the inspiration from:
https://github.com/ethereum/solc-bin/blob/gh-pages/
https://binaries.soliditylang.org/bin/list.json
https://binaries.soliditylang.org/wasm/list.json

In our case we need to expose also compatibility information between resolc and solc as our revive compiler internally calls the solc compiler

EDIT:

After the discussion the task list is:
- [ ] Change the revive release CI workflow no longer compress any releases. Meaning instead of one Wasm archive there will be 3 JS files.
- [ ] Create a `resolc-bin` repository which will host a JSON file per release platform. The structure should be same as [`solc-bin`](https://github.com/ethereum/solc-bin). [See this comment for an example](https://github.com/paritytech/revive/issues/162#issuecomment-2743087387).
- [ ] Implement the scripts and workflows required such that for each release in this repository, the JSON files are ammended with the new release.

athei:
> In our case we need to expose also compatibility information between resolc and solc as our revive compiler internally calls the solc compiler

Is that true? `resolc` should work with any `solc` version as long as it is > `0.8.0`. As long as the CLI arguments don't change.

smiasojed:
The question is whether it might change in the future, as we probably would like to avoid modifying the `list.json` format, which could already be in use by external tools.
Do we want to include information somewhere here indicating that we support Solidity versions 0.8.0 and above?

xermicus:
@smiasojed this issue might need further explanation as to why we need a website providing this information. Because the `resolc` binary will simply error out if the `solc` binary it tries to work with isn't supported (for example because it is newer than our latest supported version).

xermicus:
> `resolc` should work with any `solc` version as long as it is > `0.8.0`

That is correct as of now.

smiasojed:
> [@smiasojed](https://github.com/smiasojed) this issue might need further explanation as to why we need a website providing this information. Because the `resolc` binary will simply error out if the `solc` binary it tries to work with isn't supported (for example because it is newer than our latest supported version).

@xermicus I think that for contract tooling, we will need to download the correct Revive version along with the corresponding Solidity version to ensure a seamless configuration for developers.

[...]

PR #260, opened 2025-03-17
Add --supported-solc-versions CLI command

pkhry:
### Description 

Adds `--supported-solc-versions` cli command to return a semver range of supported `solc` versions. 
Is a hack until #162 is implemented.

### DIFF
--- a/crates/solidity/src/lib.rs
+++ b/crates/solidity/src/lib.rs
@@ -46,6 +46,9 @@ pub use self::solc::version::Version as SolcVersion;
 pub use self::solc::Compiler;
 pub use self::version::Version as ResolcVersion;
 pub use self::warning::Warning;
+
+pub use solc::{FIRST_SUPPORTED_VERSION, LAST_SUPPORTED_VERSION};
+
 #[cfg(not(target_os = "emscripten"))]
 pub mod test_utils;
 pub mod tests;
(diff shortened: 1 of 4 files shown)


crates/solidity/src/lib.rs
```diff
@@ -46,6 +46,9 @@ pub use self::solc::version::Version as SolcVersion;
 pub use self::solc::Compiler;
 pub use self::version::Version as ResolcVersion;
 pub use self::warning::Warning;
+
+pub use solc::{FIRST_SUPPORTED_VERSION, LAST_SUPPORTED_VERSION};
```
xermicus, line 50: I know it is a bit arbitrary but at least we should try to keep some consistency to not make this a complete mess.

Please export those like the rest, meaning like this (and move them up, the formatter should then sort those automatically to the correct position):

```
pub use self::solc::FIRST_SUPPORTED_VERSION as SolcFirstSupportedVersion;
pub use self::solc::LAST_SUPPORTED_VERSION as SolcLastSupportedVersion;
```

pkhry: done

xermicus:
> Is a hack until https://github.com/paritytech/revive/issues/162 is implemented.

Wait, so why aren't we just implementing #162 then?

PR #261, opened 2025-03-18
Support solc v0.8.29

xermicus:
No observable changes w.r.t. YUL compilation.
- Mark solc v0.8.29 as the latest supported version
- Add integration test for solc v0.8.29 specific feature
- Use the latest version on CI
- Drive-by fix linter complaints and a two types in the CI yaml files

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,7 @@ on:
     branches: ["main"]
   pull_request:
     branches: ["main"]
-    types: [opened, synchronize, labled, unlabled]
+    types: [opened, synchronize, labeled, unlabeled]
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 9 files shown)


xermicus:
Good catch, thanks!

.github/workflows/release.yml
```diff
@@ -4,7 +4,7 @@ on:
     branches: ["main"]
   pull_request:
     branches: ["main"]
-    types: [opened, synchronize, labled, unlabled]
+    types: [opened, synchronize, labeled, unlabeled]
```
athei, line 7: I was going mad why this shit wasn't working lol.

athei (approved):

.github/workflows/release.yml
xermicus: Yeah as obscure as it gets, I only caught it because of the VS code extension complained about it

athei: Yeah why isn't it an error to use non existent types :(

PR #262, opened 2025-03-18
remove support for webkit

xermicus:


xermicus:
@smiasojed found out that it's actually just a race condition not surfacing in the other engines.

xermicus:
Turns out it's the same issue we saw bun. We remove it for now, kicking the can down the road.

PR #269, opened 2025-03-25
ci: modify release according to #162

alvicsam:
PR modifies release workflow according to changes requested in https://github.com/paritytech/revive/issues/162#issuecomment-2751287953

cc https://github.com/paritytech/revive/issues/162
cc https://github.com/paritytech/devops/issues/3890

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -241,7 +241,7 @@ jobs:
         run: |
           lipo resolc-aarch64-apple-darwin resolc-x86_64-apple-darwin -create -output resolc-universal-apple-darwin
 
-      - name: Compress Artifacts
+      - name: Make Executable
         run: |
           chmod +x resolc-x86_64-unknown-linux-musl
           chmod +x resolc-universal-apple-darwin
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -214,15 +214,15 @@ jobs:
 
       - name: Compress Artifact
         run: |
-          tar -czf $(pwd)/resolc-wasm32-unknown-emscripten.tar.gz -C ./target/wasm32-unknown-emscripten/release/ \
-            resolc.js \
-            resolc.wasm \
-            resolc_web.js
+          mkdir -p resolc-wasm32-unknown-emscripten
+          mv ./target/wasm32-unknown-emscripten/release/resolc.js ./resolc-wasm32-unknown-emscripten/
+          mv ./target/wasm32-unknown-emscripten/release/resolc.wasm ./resolc-wasm32-unknown-emscripten/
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -214,15 +214,15 @@ jobs:
...
-            resolc.wasm \
-            resolc_web.js
+          mkdir -p resolc-wasm32-unknown-emscripten
+          mv ./target/wasm32-unknown-emscripten/release/resolc.js ./resolc-wasm32-unknown-emscripten/
+          mv ./target/wasm32-unknown-emscripten/release/resolc.wasm ./resolc-wasm32-unknown-emscripten/
+          mv ./target/wasm32-unknown-emscripten/release/resolc_web.js ./resolc-wasm32-unknown-emscripten/
 
       - uses: actions/upload-artifact@v4
         with:
           name: resolc-wasm32-unknown-emscripten
```
xermicus, line 224: Uses the same name for all three artifacts, is this fine?

.github/workflows/release.yml
```diff
@@ -214,15 +214,15 @@ jobs:
 
       - name: Compress Artifact
```
xermicus, line 215: Should be renamed to reflect what it does

.github/workflows/release.yml
alvicsam: The commands move the files to the `resolc-wasm32-unknown-emscripten` folder, they will have different names.

alvicsam:
@xermicus checked the release, the draft looks fine: https://github.com/paritytech/revive-workflow-test/releases/tag/untagged-ba2ff06d87f949b8295f
Checked 2 times, the first time the action created draft with all files including old archives, after I removed the files and run the release again it was fine (see the draft)

xermicus:
>  the first time the action created draft with all files including old archives, after I removed the files and run the release again it was fine (see the draft)

Was this a fluke? I'd rather not want to do that every time we release a new version :) Per se the workflow looks good to me, any idea why that happened?

alvicsam:
@xermicus tried one more time and there were no additional files. I can assume that it was a bug in the `softprops/action-gh-release` github action that creates the release so the old files should be removed only once.

xermicus:
Yeah I also can't see why that would happen. Let's declare it a fluke.

xermicus (approved):
Thanks!

xermicus:
Should probably keep the fork with the new release so that `resolc-bin` can be implemented against that until we have a new release here.

alvicsam:
> Should probably keep the fork with the new release so that resolc-bin can be implemented against that until we have a new release here.

Sure, I'm planning to test the json-generator script on the fork first

PR #272, opened 2025-04-03
bugfix missing byte swap for the create2 salt value

xermicus:
Found in https://github.com/paritytech/contract-issues/issues/45, thanks @albertov19 and @sekisamu


nhussein11:
@xermicus, once this is released, can it be escalated to the https://github.com/paritytech/hardhat-revive/tree/main repo so a new version of the plugin can be released with this update in the compiler?

xermicus:
@nhussein11 we will make sure to follow up with timely releases of the tooling

PR #276, opened 2025-04-04
ci: update release flow and publish list.json

alvicsam:
This pull request adds changes described in **this comment**:

- `.github/scripts/json_generator.py` - a small script that generates several json files for different platforms.
- [generate_versions.yml](https://github.com/paritytech/revive/compare/as-release-json?expand=1#diff-2aee05b96020ac60943e6dfcb30597e53898f31542aeb570468b970d9a13a5a6) - the workflow that runs when release is published, creates info.json files and pushes them into resolc-bin repo
- `js/build.js` is adjusted in order to set `RESOLC_WASM_URI` from env variable
- ⚠️ Release workflow is changed: 
  - In PRs and main branch it'll only build artifacts
  - Release will happen automatically only on the `v*` tag push. This is needed for revive_web.js to have the necessary `RESOLC_WASM_URI`
  - workflow will check that version in Cargo.toml is the same as the tag when the new tag is pushed

cc https://github.com/paritytech/revive/issues/162
cc https://github.com/paritytech/devops/issues/3890


.github/workflows/release.yml
```diff
@@ -255,9 +263,11 @@ jobs:
...
             ## Note for macOS Users
             The macOS binary is unsigned and it needs to be made runnable using `xattr -c resolc-universal-apple-darwin`.
-          tag_name: ${{ needs.check-version-changed.outputs.PKG_VER }}
-          name: ${{ needs.check-version-changed.outputs.PKG_VER }}
-          draft: true
+          tag_name: ${{ github.ref_name }}
+          name: ${{ github.ref_name }}
+          prerelease: true
+          token: ${{ steps.app-token.outputs.token }}
+          # draft: true
```
xermicus, line 270: Is the prerelease needed instead of the draft? If so please delete instead of leaving commented out.

.github/workflows/release.yml
```diff
@@ -17,41 +18,43 @@ env:
...
-      PKG_VER: ${{ steps.versions.outputs.PKG_VER }}
       RELEASE_NOTES: ${{ steps.versions.outputs.RELEASE_NOTES }}
     steps:
       - uses: actions/checkout@v4
-        with:
-          fetch-tags: true
-          fetch-depth: 0
 
-      - name: Check Versions
+      - name: Check that tag and version in Cargo.toml match
```
xermicus, line 31: Nit: Shouldn't this go into the description?

xermicus (commented):
Looking good, can you please update the `RELEASE.md` to reflect these changes?

.github/workflows/release.yml
alvicsam: What do you mean? Neither jobs nor steps don't have a description. I can leave a comment before the step.

alvicsam:
@xermicus thanks for the review! Please check the updated `Release.md`

.github/workflows/release.yml
xermicus: What I mean is that this reads like a description now while it is supposed to be a title

xermicus: A comment would also be fine

xermicus (approved):
Thanks!

Issue #266, opened 2025-03-23
Allocating an array of length 63 in emit causes execution to fail.


Subway2023:
Resolc version: 0.1.0-dev.12+commit.76f4cf7.llvm-18.1.8
Description: **emit ev(42, new uint16 );** causes execution to fail. However, following the commented method does not result in failure. Both solc+evm and zksolc+zkvm execute successfully.

```solidity
contract C {
    event ev(uint8, uint16[]);

    bytes s;

    function f() public {
        emit ev(42, new uint16[](63));

        // uint[] memory dynamicArray;
        // dynamicArray = new uint[](63);
    }
}
```
## Run with resolc
I have set up an execution [environment ](https://github.com/Subway2023/revive)that can directly execute the bytecode generated by revive compilation.
### compile
```
resolc --solc solc test.sol --combined-json hashes,bin,abi -o /tmp/tmpdh_0t_pr --overwrite
```
### execute
```
[...]

PR #284, opened 2025-04-22
revive-runner: add a utility binary for local contract execution

xermicus:
I had this in mind for a while but never implemented a standalone binary so far because I always end up writing an integration test anyways.

However, using a standalone version of the pallet based on the revive-runner crate is something people filing in bug reports do anyways, for example:
https://github.com/paritytech/revive/issues/266
https://github.com/paritytech/contract-issues/issues/54

I hope having this maintained in the upstream code base here helps people with desired functionality and also keeping the version up to date. @Subway2023 @sekisamu just hit me up here or open an issue or PR here for any missing features you'd like.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This is a development pre-release.
 Supported `polkadot-sdk` rev:`c29e72a8628835e34deb6aa7db9a78a2e4eabcee`
 
 ### Added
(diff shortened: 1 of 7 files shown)


sekisamu:
@xermicus  Currently, the stack size of a Polkavm instance is determined by the contract bytecode. It would be super useful if we could customize it during debugging

PR #292, opened 2025-04-26
add homebrew taps

mikelxc:
# Add Homebrew Formula for Revive Solidity Compiler (resolc)

This PR adds a Homebrew Formula for the Revive Solidity Compiler (resolc), making it easily installable on macOS systems.

## Changes

- Added `Formula/resolc.rb` with support for binary installation
- Added GitHub workflow to automatically update the Formula on new releases
- Includes necessary steps to handle unsigned macOS binaries using `xattr -c`

## Installation

Once merged, users can install resolc using:
```bash
brew tap paritytech/revive
brew install resolc
```

## Notes

- The Formula uses the pre-built binary from GitHub releases
- Includes automatic SHA256 verification for security
- The GitHub workflow will automatically update the Formula when new releases are published

## Testing

The Formula has been tested with the latest release (v0.1.0-dev.14) and includes a basic test to verify the installation.

### DIFF
--- a/.github/workflows/update-formula.yml
+++ b/.github/workflows/update-formula.yml
@@ -1,0 +1,43 @@
+name: Update Homebrew Formula
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  update-formula:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.2'
+
+      - name: Install dependencies
+        run: |
+          gem install bundler
+          bundle install
+
+      - name: Update Formula
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Get the release version and SHA256
+          VERSION=$(echo ${{ github.ref }} | sed 's/refs\/tags\/v//')
+          SHA256=$(curl -sL https://github.com/paritytech/revive/archive/refs/tags/v${VERSION}.tar.gz | shasum -a 256 | cut -d' ' -f1)
+          
+          # Update the Formula file
+          sed -i "s/url .*/url \"https:\/\/github.com\/paritytech\/revive\/archive\/refs\/tags\/v${VERSION}.tar.gz\"/" Formula/resolc.rb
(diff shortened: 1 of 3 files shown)


Formula/README.md
```diff
@@ -0,0 +1,28 @@
...
+brew tap paritytech/revive
+brew install resolc
+```
+
+## Updating
+
+To update the formula:
+
+1. Create a new release on GitHub
+2. Update the `url` and `sha256` in `Formula/resolc.rb`
```
athei, line 19: Stale comment? This should be handled by the github workflow?

Formula/resolc.rb
```diff
@@ -0,0 +1,33 @@
+class Resolc < Formula
+desc "Revive Solidity Compiler"
+homepage "https://github.com/paritytech/revive"
+url "https://github.com/paritytech/revive/releases/download/v0.1.0-dev.14/resolc-universal-apple-darwin"
+sha256 "7d5b3cd4233e60e8bfaade398062f1b609166dcb251424cd9d402e51126d8b4f"
+license "MIT/Apache-2.0"
+  
+# If someone wants to build from latest source:
+head "https://github.com/paritytech/revive.git", branch: "master"
+depends_on "llvm@18" => :build if build.head?    # only needed when building from source
```
athei, line 10: This won't work because we need our custom LLVM in order to build resolc. So either remove or add a second Formula for our LLVM build. We also do binary releases of those in our repo.

.github/workflows/update-formula.yml
```diff
@@ -0,0 +1,43 @@
+name: Update Homebrew Formula
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  update-formula:
+    runs-on: ubuntu-latest
```
athei, line 9:
```suggestion
    runs-on: ubuntu-24.04
```

We use fixed version by convention.

.github/workflows/update-formula.yml
```diff
@@ -0,0 +1,43 @@
...
+          gem install bundler
+          bundle install
+
+      - name: Update Formula
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # Get the release version and SHA256
+          VERSION=$(echo ${{ github.ref }} | sed 's/refs\/tags\/v//')
+          SHA256=$(curl -sL https://github.com/paritytech/revive/archive/refs/tags/v${VERSION}.tar.gz | shasum -a 256 | cut -d' ' -f1)
```
athei, line 32: We also do LLVM releases from this repo. So we need to detect this here and skip if it is a LLVM release.

PR #298, opened 2025-05-05
apply size optimizations by default

xermicus:
So far if no optimization level was specified, optimizations for execution time were applied. However, we currently are a bit limited on code size. Add to that, this setting is not available in solc and people generally ignore the docs, generating a lot of support requests.


xermicus:
That CI fail is odd. Apparently this did not surface so far. But it seems like the builtins archive on the windows build is corrupt.

PR #308, opened 2025-05-08
support solc `v0.8.30`

xermicus:
- Add support for solc `v0.8.30`. No changes in YUL or the binary interface.
- Fix a semver bug in the NPM packages to correctly align their solc dependencies with our last supported version.


js/resolc/package.json
```diff
@@ -32,6 +32,6 @@
         "@types/node": "^22.9.0",
         "commander": "^13.1.0",
         "package-json": "^10.0.1",
-        "solc": "^0.8.29"
+        "solc": "<=0.8.30"
```
athei, line 35: Shouldn't we also require a minimum version here?

xermicus: Yes

athei (approved):

js/resolc/package.json
xermicus: Restricted it further. Good catch thanks!

PR #310, opened 2025-05-08
Add compiler optimizer option

pgherveou:
- **Run lint:fix**
- **add optimizer options:**

### DIFF
--- a/js/resolc/src/index.ts
+++ b/js/resolc/src/index.ts
@@ -5,114 +5,114 @@ import path from 'path'
 import { existsSync, readFileSync } from 'fs'
 
 export type SolcInput = {
-    [contractName: string]: {
-        content: string
-    }
+  [contractName: string]: {
+    content: string
+  }
 }
 
 export type SolcError = {
-    component: string
-    errorCode: string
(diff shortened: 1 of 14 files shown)


js/resolc/src/index.ts
```diff
@@ -93,13 +93,26 @@ export function version(): string {
...
   sources: SolcInput,
-  option: { bin?: string } = {}
+  option: {
+    optimizer?: Record<string, unknown>
+    bin?: string
+  } = {}
 ): Promise<SolcOutput> {
+  const {
+    optimizer = {
+      mode: 'z',
```
xermicus, line 103: Should this as well as the optimizer runs be argumenta to the `compile` functions?

xermicus (approved):

js/resolc/src/index.ts
pgherveou: not sure what you mean. if no default is specified we will use that, else what ever the user specify

xermicus: Ah yes of course, I thought it would not apply to the bin but it does. Nevermind

PR #318, opened 2025-05-12
expose custom PVM settings in the standard json interface

### LINKED ISSUE #290: Support extra arguments via STD  json
For tooling it is easier to pass extra options (which don't exist in solc) via std or combined json input modes too. This is already implemented for the LLVM optimization level with an augmentation of the solc std json optimizer settings.

Candidates:
- Stack size
- Heap size
- Debug info

xermicus:
Exposes the following PolkaVM specific options via the standard json interface:
- Heap size
- Stack size
- Whether to emit source level debug information

Additionally it is now forbidden to specify those as CLI option in standard JSON mode. They are bytecode altering options and having multiple ways to specify them creates unnecessary room for confusion: The standard JSON input description should be sufficient and succint for reproducible builds.

Closes #290


athei:
Nice. Agreed. Having multiple ways of doing the same thing is usually a bad sign.

PR #337, opened 2025-05-30
solc-json-interface: do not uncoditionally skip serialization of custom keys

xermicus:


crates/solc-json-interface/src/standard_json/input/settings/polkavm/mod.rs
```diff
@@ -11,16 +11,18 @@ pub mod memory;
 #[derive(Clone, Copy, Default, Debug, Serialize, Deserialize)]
 pub struct PolkaVM {
     /// The PolkaVM target machine memory configuration settings.
-    pub memory_config: MemoryConfig,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
smiasojed, line 14: Would it be possible to add `#[serde(default, skip_serializing_if = "Option::is_none")]` to each struct field? That way, I could use the struct directly in Foundry.

xermicus: Isn't that futile? Serde deserializes absent optional fields to `None` anyways.

Issue #320, opened 2025-05-16
Do not request EVM bytecode outside test


xermicus:
I noticed we always request the EVM bytecode from solc. We need it during tests and benchmarks. But it is unnecessary when compiling to PVM. This should be changed because emitting the bytecode is extra work in solc. Instead, the test utils should just request it explicitly and then resolc adds the EVM runtime and deploy codes to the standard JSON settings supplied to solc.

PR #338, opened 2025-05-30
fix: exclude EVM bytecode from production solc requests

0xf333:
This is to address issue #320

## Introduced changes
- Added new_required_for_tests() method that includes EVM bytecode flags
- Modified new_required() to exclude evm.bytecode and evm.deployedBytecode
- Updated test utilities to explicitly request EVM bytecode when needed


xermicus (approved):
Thanks!

Can you please sign off your commit (`git commit --amend --signoff` might do it if you configured code signing already).

Issue #355, opened 2025-06-30
Nightly builds


xermicus:
I want to have nightly builds of the `main` branch (given there was at least one commit since the last nightly build).

This should be re-using the logic from the existing release process - however if possible I'd like to not create a GH release for each nightly build but just publish them to [resolc-bin](https://github.com/paritytech/resolc-bin) directly.

alvicsam:
Out of curiosity: why don't you want to have them in releases in this repo? If not how do you see that? Pushing binaries directly into https://github.com/paritytech/resolc-bin? In this case there is a problem with git, the repo size is going to be enormous and slow in a couple of months.

We also could create nightly releases in the [resolc-bin](https://github.com/paritytech/resolc-bin) repo. We don't need to append them to the json files, right?

xermicus:
Intuitively I just feel like having nightly releases will spam the release page. Users will end up scrolling endlessly to find any older one. 

Just checked [solc](https://github.com/ethereum/solidity/releases), they also don't do it. Instead they  push to their binaries repo. I thought we could do the same?

> We don't need to append them to the json files, right?

I don't see them for example [here](https://binaries.soliditylang.org/wasm/list.json) so I guess no. CC @alvicsam

alvicsam:
Yes, but on the other hand they don't put binaries of nightly releases into the solc-bin repo. You can see that for example in the `linux-amd64` folder they have about 90 binaries for the **5 years**. I tried a git clone of the repo and after 6 minutes it could download only 24%. Revive has 340 MB of binaries on each release. You can image how bad it's going to be in 3 months if we put binaries into git. Also [the ci workflow that generates json files](https://github.com/paritytech/revive/blob/main/.github/workflows/generate_versions.yml) is going to stop working at some point because it checks out the resolc-bin repo.

alvicsam:
If the nightly builds needed only for small amount of time we can have a nightly pipeline that will build artifacts and keep them for a few days. According to github [docs](https://docs.github.com/en/organizations/managing-organization-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-organization) they can be kept for 90 days. 
Another solution would be creating a `revive-nightly-releases` repo and do the nightly releases there.

smiasojed:
Why do we need to push binaries to the `resolc-bin` repo? Don’t we already reference binaries via JSON files pointing to releases in the `revive` repo?
[...]

PR #357, opened 2025-07-03
Generate nightly builds and json files

alvicsam:
PR adds a nightly release pipeline.
Since the logic for the nightly build and release differs significantly from the usual pipeline, I decided to put it in a separate file. The same applies to the Python script that generates JSON files.

The nightly pipeline runs at 01:00 UTC every day, producing artefacts and retaining them for 40 days if commits are made the day before it runs.

cc https://github.com/paritytech/revive/issues/355
cc https://github.com/paritytech/devops/issues/4141

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,12 +2,11 @@ name: Build & Release
 on:
   push:
     branches: ["main"]
-  # temporary disable for tests
-  #   tags:
-  #     - "v*"
-  # pull_request:
-  #   branches: ["main"]
-  #   types: [opened, synchronize, labeled, unlabeled]
+    tags:
+      - "v*"
+  pull_request:
+    branches: ["main"]
+    types: [opened, synchronize, labeled, unlabeled]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -14,6 +14,7 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
+  # if changed, dont forget to update the env var in release-nightly.yml
```
xermicus, line 17: In theory the nightly release should just have different commit inputs right? Then no actual build logic needs to be duplicated. I see a lot of duplicated code in `release-nightly.yml` - for maintainability I'd appreciate making release.yml more modular (maybe as it's own GH actions or even a couple ones). WDYT?

xermicus (commented):
Thank you @alvicsam!

.github/workflows/release.yml
alvicsam: > In theory the nightly release should just have different commit inputs right? 

In theory, yes, and the build jobs are the same, but the logic for working with artifacts differs. I tried to keep everything in one pipeline. However, there were too many rules about whether a job or step should run, which made the pipeline definition difficult to read.

We could merge this one so it starts producing nightly builds and do the refactor later. WDYT?

xermicus: Maybe the rules could be modularized too? 

I am fine with merging it to simplify trying it out - if you promise to improve it afterwards :)

alvicsam: AFAIK no, it's not applicable to the rules. I'll create a separate issue then for the refactor.

PR #372, opened 2025-08-15
Implement Rust CLI tests

### LINKED ISSUE #365: Convert CLI tests from TS to Rust
Having a split test suite (TS and Rust) is a major maintenance burden and there is no reason they have to be written in TS.

- [ ] https://github.com/paritytech/revive/tree/main/crates/resolc/src/tests/cli-tests
- [ ] https://github.com/paritytech/revive/blob/main/crates/resolc/src/tests/combined-json.tests.ts

elle-j:
# Description

Closes #365 

Ports all `resolc` CLI tests to Rust and removes the CLI TypeScript tests.

### DIFF
--- a/crates/resolc/src/tests/cli/optimization.rs
+++ b/crates/resolc/src/tests/cli/optimization.rs
@@ -34,3 +34,20 @@ fn runs_with_valid_level() {
         );
     }
 }
+
+#[test]
+fn fails_with_invalid_level() {
+    let arguments = &[utils::YUL_MEMSET_CONTRACT_PATH, yul::YUL_OPTION, "-O9"];
+    let resolc_result = utils::execute_resolc(arguments);
+    utils::assert_command_failure(&resolc_result, "Providing an invalid optimization level");
+
+    assert!(
+        resolc_result
+            .stderr
+            .contains("Unexpected optimization option"),
+        "Expected the output to contain a specific error message."
+    );
+
+    let solc_result = utils::execute_solc(arguments);
+    utils::assert_equal_exit_codes(&solc_result, &resolc_result);
+}


Makefile
```diff
@@ -68,9 +67,6 @@ test-resolc: install
 test-workspace: install
 	cargo test --workspace --exclude revive-llvm-builder
 
-test-cli: install
-	npm run test:cli
```
elle-j, line 72: We could also keep the `make test-cli` and have it run the Rust tests (perhaps with `cargo test tests::cli`), in order to keep the short and sweet way of running only those.

crates/resolc/src/tests/cli/yul.rs
```diff
@@ -0,0 +1,44 @@
+//! The tests for running resolc with yul option.
+
+#![cfg(test)]
+
+use crate::tests::cli::utils;
+
+pub const YUL_OPTION: &str = "--yul";
+// The `--yul` option was deprecated in Solidity 0.8.27 in favor of `--strict-assembly`.
+// See section `--strict-assembly vs. --yul` in https://soliditylang.org/blog/2024/09/04/solidity-0.8.27-release-announcement/
```
xermicus: This should be a doc-comment

```suggestion
/// The `--yul` option was deprecated in Solidity 0.8.27 in favor of `--strict-assembly`.
/// See section `--strict-assembly vs. --yul` in https://soliditylang.org/blog/2024/09/04/solidity-0.8.27-release-announcement/
```

crates/resolc/src/tests/cli/asm.rs
```diff
@@ -0,0 +1,40 @@
...
+
+#![cfg(test)]
+
+use crate::tests::cli::utils;
+
+const ASM_OPTION: &str = "--asm";
+
+#[test]
+fn runs_with_valid_input_file() {
+    const ARGUMENTS: &[&str] = &[utils::SOLIDITY_CONTRACT_PATH, ASM_OPTION];
```
xermicus: Nit: We usually don't do this; bindings in Rust are immutable anyways unless they are explicitly made mutable via the `mut` keyword. It's more common to just bind those parameters to a binding with `let`. Then you also don't need to provide the type signature and snake case variable names are a bit more pleasant to read than those uppercase constants.

I see in other tests sometimes you actually did it with bindings but sometimes like this. I'd appreciate if you could change to just have bindings everywhere.

crates/resolc/src/tests/cli/utils.rs
```diff
@@ -0,0 +1,103 @@
...
+pub const YUL_CONTRACT_PATH: &str = "src/tests/cli/contracts/yul/contract.yul";
+pub const YUL_MEMSET_CONTRACT_PATH: &str = "src/tests/cli/contracts/yul/memset.yul";
+pub const STANDARD_JSON_CONTRACTS_PATH: &str =
+    "src/tests/cli/contracts/standard_json/solidity_contracts.json";
+
+/// The result of executing a command.
+pub struct CommandResult {
+    /// The data written to `stdout` if any,
+    /// otherwise, the data written to `stderr`.
+    pub output: String,
```
xermicus: I find this a bit weird; we're unable to tell apart whether the output came from stdout or stderr. Also processes can write to both pipes. What about having two struct fields, `stdout` and `stderr`, instead?

Cargo.toml
```diff
@@ -71,6 +71,7 @@ tar = "0.4"
 toml = "0.8"
 assert_cmd = "2.0"
 assert_fs = "1.1"
+rstest = "0.26.1"
```
xermicus: I don't like this crate because it introduces proc macros for very simple things that can be expressed in a very few lines of extra code. People think it's more elegant because it might sometimes result in less code and this is therefor better. I think this is a fallacy and for maintainability it's generally better to express things in pure Rust and not let others reverse engineer (or read docs on) proc-macros. Unless a lot of code is generated or the proc-macro is very popular (for example serde), which doesn't seem to be the case here.

But it's also not the end of the world, I'll leave it up to you if you want to keep the crate or rewrite the few places where it's used.

crates/resolc/Cargo.toml
```diff
@@ -36,6 +36,9 @@ revive-llvm-context = { workspace = true }
 revive-solc-json-interface = { workspace = true, features = ["resolc"] }
 revive-yul = { workspace = true }
 
+[dev-dependencies]
+rstest = { workspace = true }
```
xermicus:
```suggestion
rstest = { workspace = true, default_features = false }
```

Makefile
xermicus: Removing this is fine for me, I think the split was more naturally when they were not executed with `cargo test`.

crates/resolc/src/tests/cli/asm.rs
elle-j: Good to know the convention here, thanks! 🙂 I used `const` for where it was possible, and `let` elsewhere. I'd prefer the more easily readable one as well

crates/resolc/src/tests/cli/utils.rs
elle-j: This was mainly to retain the same behavior as before, but we could definitely update this. Replacing the `output` with the two other fields is a good suggestion 👍

Cargo.toml
elle-j: Thanks for the insight. I'll go ahead and see what a refactor will look like

xermicus: Maybe I am getting old and out of touch. But I think now the tests are simpler to read with barely any additional lines of code added. Thanks!

crates/resolc/src/tests/cli/utils.rs
xermicus: Cool, looks much better now thanks

xermicus (approved):
Thanks!

PR #387, opened 2025-10-07
Feat/compiler explorer UI 366

gboigwe:


xermicus:
Closing this as we decided to integrate resolc with the godbolt.org compiler explorer instead.

PR #390, opened 2025-10-17
Strip `minsize` attribute from functions with large div/rem

kvpanch:
That's a workaround to avoid LLVM backend crash while selecting instruction for 256-bit integer division.
The workaround needs to be removed after we switch to newest inkwell that has a fix in RISC-V backend


kvpanch:
@xermicus looks like there's no tests to verify successful compilation only, right ?

crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -1394,4 +1406,43 @@ impl<'ctx> Context<'ctx> {
...
+                        InstructionOpcode::SDiv
+                        | InstructionOpcode::UDiv
+                        | InstructionOpcode::SRem
+                        | InstructionOpcode::URem =>
+                        // This might be too aggressive, because compiler works only crashes for
+                        // power-of-2 cases, otherwise works correctly.
+                        // However, if the second operand is not a constant at this moment, there's
+                        // a small change ExpandDivRemPass will still fail if some conprop
+                        // optimization will be done.
+                        return inst.get_type().into_int_type().get_bit_width() >= 256,
```
kvpanch: This can be slightly relaxed to ignore non-power-of-2 constant values, but I've struggled to understand how to get integer value from ConstantInt using inkwell. Should be able to find it later (assuming it's possible)

xermicus: I tried that before but it isn't as trivial as it seems. IIRC this needs to be done after LLVM constant folding because YUL isn't always properly constant folded (likely cause is the YUL optimizer doing this explicitly to make for smaller code size on EVM). I thought that instead of removing the attribute we could change the operation to a shift. But let's not change the LL code after opt again. Removing the attribute for any divisions above certain bit size like you did is fine for me.

crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -1394,4 +1406,43 @@ impl<'ctx> Context<'ctx> {
...
+                    match inst.get_opcode() {
+                        InstructionOpcode::SDiv
+                        | InstructionOpcode::UDiv
+                        | InstructionOpcode::SRem
+                        | InstructionOpcode::URem =>
+                        // This might be too aggressive, because compiler works only crashes for
+                        // power-of-2 cases, otherwise works correctly.
+                        // However, if the second operand is not a constant at this moment, there's
+                        // a small change ExpandDivRemPass will still fail if some conprop
+                        // optimization will be done.
```
xermicus: Can you move this to the functions doc comment?

crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -1394,4 +1406,43 @@ impl<'ctx> Context<'ctx> {
...
+                        return inst.get_type().into_int_type().get_bit_width() >= 256,
+                        _ => return false
```
xermicus:
```suggestion
                        _ => false
```

`make test` will run the linter and formatter to catch this (should make CI green). I just let the formatter run on writes in vim.

crates/llvm-context/src/polkavm/context/mod.rs
xermicus:
```suggestion
                        inst.get_type().into_int_type().get_bit_width() >= 256,
```

xermicus (approved):
Thanks! Tried this branch on the balancer contracts where we found the bug. `-Oz` works again there and switching from `Os` to `Oz` also halves the size of the `Vault.sol` contract blob (from `574K` to `287K`)! \o/ I wouldn't mind if you could add a simple cli test for it though so we don't have to check it manually once we upgrade to LLVM 21.

crates/llvm-context/src/polkavm/context/mod.rs
kvpanch: To clarify: I want to update logic with "if denominator is known constant and is not power-of-2, continue; remove attribute otherwise". In this case we won't strip that attribute unnecessary. 

> I tried that before but it isn't as trivial as it seems. IIRC this needs to be done after LLVM constant folding because YUL isn't always properly constant folded (likely cause is the YUL optimizer doing this explicitly to make for smaller code size on EVM)

Do you mean you saw problems with that approach too ?

xermicus: Ah what I meant was it doesn't work before LLVM optimizations. Yeah you can do that - it's annoying because for larger values there's no typed API and we can only extract it in string representation:

```rust
let string_value = inst
    .get_operand(1)
    .unwrap()
    .left()
    .unwrap()
    .into_int_value()
    .print_to_string()
    .to_string();
```

Which will emit something like for example `i256 1000000000000000000` - so I guess whatever the LLVM value to string printer emits there needs to be parsed back into a num::BigInt and checked if exactly a single bit is set.

kvpanch: argh... I stuck to understand how to access to `get_operand`'s result value. Annoyingly for me it's not easy to find good examples in inkwell's tests.

>  so I guess whatever the LLVM value to string printer emits there needs to be parsed back into a num::BigInt and checked if exactly a single bit is set.

That's fair. I was wrongly hoping `IntValue` has access to `APInt` that has convenient method to check if it's power-of-2 or not. But neither C wrappers, nor inkwell provide access to Constant's `getValue`

That said, I have to abandon that idea for now.

xermicus: Yeah it's sometimes not obvious - there's instruction types, int value type, basic value enum, the `Either` return type on the operand value etc can obscure it a bit - also `print_to_string` isn't obvious. ChatGPT is usually good for finding the APIs - here it thought for a solid minute :)

> That said, I have to abandon that idea for now.

Fair. I agree. We can merge without making it perfect. It's a workaround anyways.

PR #346, opened 2025-06-16
feat: add LLVM as git submodule

0xf333:
## Introduced changes
- Replaced revive-llvm clone with git submodule approach.
- Pin LLVM to commit `3b5b5c1ec4a3095ab096dd780e84d7ab81f3d7ff (llvmorg-18.1.8)`.
- Updated Makefile to use 'git submodule update --init --recursive'.
- Modified `.gitignore` to allow llvm submodule directory.
- Removed `LLVM.lock` file and lock mechanism from llvm-builder.
- Replaced `clone/checkout` commands with submodule initialization.
- Updated `CI` workflows to use submodules: `true` in checkout actions.
- Simplified `revive-llvm` CLI to build/clean/builtins commands only.

Closes issue  #306

PR #399, opened 2025-10-31
add LLVM 18.x as a git submodule

kvpanch:
This changeset is based on https://github.com/paritytech/revive/pull/346

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ This is a development pre-release.
 
 Supported `polkadot-sdk` rev: `2509.0.0`
 
+### Changed
+- Instruct the LLVM backend and linker to `--relax` (may lead to smaller contract code size).
+- Standard JSON mode: Don't forward EVM bytecode related output selections to solc.
+
+### Fixed:
+- The missing `STOP` instruction at the end of `code` blocks.
+
 ## v0.5.0
 
 This is a development pre-release.
--- a/crates/integration/codesize.json
+++ b/crates/integration/codesize.json
@@ -1,10 +1,10 @@
 {
-  "Baseline": 914,
-  "Computation": 2295,
-  "DivisionArithmetics": 14496,
-  "ERC20": 17482,
-  "Events": 1674,
-  "FibonacciIterative": 1490,
-  "Flipper": 2086,
-  "SHA1": 8158
+  "Baseline": 905,
+  "Computation": 2286,
+  "DivisionArithmetics": 14347,
+  "ERC20": 16929,
+  "Events": 1665,
+  "FibonacciIterative": 1447,
+  "Flipper": 2077,
+  "SHA1": 7721
 }
--- a/crates/integration/src/tests.rs
+++ b/crates/integration/src/tests.rs
@@ -1,6 +1,7 @@
 use std::str::FromStr;
 
 use alloy_primitives::*;
+use resolc::test_utils::build_yul;
 use revive_runner::*;
 use SpecsAction::*;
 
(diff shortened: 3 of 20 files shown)


.github/workflows/test-llvm-builder.yml
```diff
@@ -4,7 +4,6 @@ on:
     branches: ["main"]
     types: [opened, synchronize]
     paths:
-      - 'LLVM.lock'
```
xermicus, line 7: Maybe this should run when the `.gitmodules` has changed?

.github/workflows/release.yml
```diff
@@ -27,6 +27,8 @@ jobs:
       RELEASE_NOTES: ${{ steps.versions.outputs.RELEASE_NOTES }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
```
xermicus: We aren't building LLVM during the regular release so this shouldn't be required?

.github/workflows/release.yml
```diff
@@ -130,6 +132,8 @@ jobs:
     environment: tags
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
```
xermicus: Same here

.github/workflows/generate_versions.yml
xermicus, line 1: Also here we shouldn't require the LLVM submodule or am I missing something?

.github/workflows/test.yml
```diff
@@ -22,6 +22,8 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
```
xermicus: Same here, shouldn't be needed?

.github/workflows/test-wasm.yml
```diff
@@ -26,6 +26,8 @@ jobs:
         shell: bash
     steps:
       - uses: actions/checkout@v4
+        with:
+          submodules: true
```
xermicus: Same here

xermicus (commented):
Oh no - the LLVM builder workflow not showing up is due to an invalid spec: https://github.com/paritytech/revive/actions/runs/19039292406

This isn't this PR's fault but would you mind fixing it here at the same time as drive-by?

.github/workflows/generate_versions.yml
kvpanch: will remove

.github/workflows/test-llvm-builder.yml
kvpanch: The file no longer exists. The way it suppose to work in future is we update new `llvm` file with new commit to pull from LLVM

.github/workflows/test.yml
kvpanch: let me try to remove it

crates/llvm-builder/tests/build.rs
xermicus, line 1: I think removing the `clone` command (line 143) will fix the build

kvpanch: Let me try that.
The biggest problem is I have no linux machine to do local debugging. Do we have a chance to get remote access to any machine with linux ?

xermicus: Docker on macos is an easy option see for example the `Dockerfile` which does a musl build

On another thought maybe we may want to add the emsdk as a submodule too wdyt? That would kind of move the implicit deps from the llvm builder and CI scripts.

I got linux hosts at hand can also check this and make a piggyback PR.

xermicus:
Yeah debugging those LLVM builds can be very time consuming.

xermicus (approved):
Looks very clean. Thanks!

Issue #379, opened 2025-09-29
Support the `IPFS` hash type


xermicus:
Should be implemented same as in solc:
- Check if PVM bytecode format allows for appending arbitrary data (presumably the case - implement the necessary change in PolkaVM if necessary).
- Implement the feature:
  - Generate the IPFS hash
  - Append the CBOR header
- Implement basic tests that ensure ["Perfect verification"](https://docs.soliditylang.org/en/latest/060-breaking-changes.html#metadata-hash-options)

PR #402, opened 2025-11-03
feat: added support for the IPFS hash type

0xf333:
@xermicus 

this is to address issue #379 

## implementation test
```bash
cargo test -p resolc ipfs
```
```bash
cargo test -p revive-solc-json-interface ipfs
```


xermicus (changes requested):
Hi, many thanks for the PR!

The end goal (to make this useful first hand) is to append it as CBOR data to the bytecode.

See also the `solc` documentation:
https://docs.soliditylang.org/en/latest/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode
https://docs.soliditylang.org/en/latest/060-breaking-changes.html#metadata-hash-options

Sorry it doesn't say it in the issue. I added a few sentences to the issue.

Issue #403, opened 2025-11-06
Benchmark resolc compilation time


elle-j:
Add compilation time benchmarks for resolc end-to-end.

Issue #199, opened 2025-02-06
Execute compilation benchmarks on CI


xermicus:
We need a CI workflow which benchmark the compilation time and memory consumption of the compiler on our reference bench-marking hardware.

Note that currently there are no such benchmarks implemented. So the issue has two tasks:
- [ ] Implement such benchmarks.
- [ ] Implement a CI workflow executing them.

I'd like to have this workflow executed after all other jobs finished and post the benchmark results (diffed against master) as a PR comment.

A follow up issue will be to create more benchmarks that focus on more realistic examples from our overarching EVM test suite.

PR #406, opened 2025-11-06
Implement `resolc` end-to-end compilation benchmarks

elle-j:
# Description

Closes [#403](https://github.com/paritytech/revive/issues/403)

Adds compilation time benchmarks for resolc end-to-end.

The benchmarks can be run from the root via:

```sh
make bench-resolc
```

HTML reports will be generated under `target/criterion`, and a summary of the results at [crates/resolc/BENCHMARKS_M4PRO.md](https://github.com/paritytech/revive/blob/lj/compilation-benchmarks/crates/resolc/BENCHMARKS_M4PRO.md) (currently from running on a Mac M4 Pro).

### DIFF
--- a/crates/resolc/BENCHMARKS_M4PRO.md
+++ b/crates/resolc/BENCHMARKS_M4PRO.md
@@ -50,3 +50,4 @@
 
 ---
 Made with [criterion-table](https://github.com/nu11ptr/criterion-table)
+
--- a/crates/resolc/src/tests/cli/bin.rs
+++ b/crates/resolc/src/tests/cli/bin.rs
@@ -9,7 +9,7 @@ use crate::{
     SolcCompiler,
 };
 
-/// The starting hex value of a PVM blob (encoding of `"PVM"`)).
+/// The starting hex value of a PVM blob (encoding of `"PVM"`).
 const PVM_BLOB_START: &str = "50564d";
 
 /// The starting hex value of an EVM blob compiled from Solidity.


crates/resolc/benches/compile.rs
```diff
@@ -0,0 +1,116 @@
...
+    });
+
+    group.finish();
+}
+
+fn bench_baseline(c: &mut Criterion) {
+    let name = "Baseline";
+    let group = group(c, name);
+    let path = get_contract_path(name);
+    let compiler_arguments = &[&path, "-O0"];
```
kvpanch: can you please clarify why `O0` optset is used ?

crates/resolc/BENCHMARKS_M4PRO.md
```diff
@@ -0,0 +1,46 @@
+# Benchmarks
```
kvpanch, line 1: If I understand it right, this file currently contains data from your run (m4 pro).  Is it expected that developers will updated that file each commit/important commit ?

crates/resolc/benches/compile.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+fn get_stderr(result: &Output) -> String {
+    String::from_utf8_lossy(&result.stderr).to_string()
+}
+
+/// Get the relative path to a `.sol` contract file with file stem `name`.
+/// The file is expected to live under `"crates/integration/contracts/"`.
+fn get_contract_path(name: &str) -> String {
+    format!("crates/integration/contracts/{name}.sol")
```
xermicus: This seems brittle (only works inside the workspace). It's also odd to depend on the `revive-integration` fixtures but then implement the compilation part manually.

However, we don't even need `revive-integration` here: Please just use the fixtures from the tests of this crate.

Additionally, below cases benchmark mostly the same (small Solidity contracts). Would you mind to also add `--yul` and `--standard-json` benchmarks? The required fixtures of those are already there so this should be very little extra work.

crates/resolc/benches/compile.rs
```diff
@@ -0,0 +1,116 @@
...
+fn group<'error, M>(c: &'error mut Criterion<M>, group_name: &str) -> BenchmarkGroup<'error, M>
+where
+    M: Measurement,
+{
+    c.benchmark_group(group_name)
+}
+
+fn bench(mut group: BenchmarkGroup<'_, WallTime>, compiler_arguments: &[&str]) {
+    group.measurement_time(Duration::from_secs(10));
+    group.sample_size(10);
```
xermicus: Why did you pick this configuration?

10s seems very high, like in the ballpark of production dApps. The sample size however is quite small. Unless it's absolutely necessary please decrease the target measurement time and increase the sample size (or don't set it if the default doesn't emit any warnings).

crates/resolc/BENCHMARKS.md
```diff
@@ -0,0 +1,46 @@
...
+    - [ERC20](#erc20)
+    - [SHA1](#sha1)
+    - [Storage](#storage)
+    - [Transfer](#transfer)
+
+## Benchmark Results
+
+### Baseline
+
+|        | `Resolc`                   |
```
xermicus: Can we add `solc` too? Otherwise I think there's little benefit of emitting these tables.

crates/resolc/benches/compile.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    group.bench_function("Resolc", |b| {
```
xermicus:
```suggestion
    group.bench_function("resolc", |b| {
```

crates/resolc/Cargo.toml
```diff
@@ -48,8 +48,14 @@ git2 = { workspace = true, default-features = false }
 
 [dev-dependencies]
 tempfile = { workspace = true }
+criterion = { workspace = true }
+revive-integration = { workspace = true }
 
 [features]
 parallel = ["rayon", "revive-solc-json-interface/parallel"]
 default = ["parallel"]
+bench-resolc = []
```
xermicus: Why is this feature flag needed?

xermicus (changes requested):

crates/resolc/benches/compile.rs
xermicus: Very good point. I think the idea is to minimize the impact of LLVM optimizations as that's outside of our control.

However the same can be said about `solc`.

So for E2E benchmarks like this I think it might be sensible to specify `Oz` or `O3` (and enable full optimizations in the comparing `solc` benchmarks).

crates/resolc/BENCHMARKS_M4PRO.md
xermicus: Fine for me if this is being vetted manually. I don't think it will change often.

But maybe there's a criterion option to include basic system configuration on which the benchmark was run on?

crates/resolc/benches/compile.rs
elle-j: We previously [talked about](https://github.com/paritytech/revive/issues/199#issuecomment-3456516744) using `-O0`, but yeah, could be more meaningful to have either all third-party optimzations disabled (solc and LLVM) or just use `-O3`. I can then add benchmarks for `solc` for comparison with optimzations as suggested, and have optimizations enabled for both 👍

crates/resolc/BENCHMARKS_M4PRO.md
elle-j: From the current [PVM/EVM execution BENCHMARKS.md](https://github.com/paritytech/revive/blob/main/crates/benchmarks/BENCHMARKS.md) my interpretation has been that it doesn't necessarily get continuously updated. When I've run those benchmarks I haven't e.g. committed the updated results.

For this PR I committed it to both keep it consistent with the other existing benchmarks, but also for reviewers to quickly see what was being produced.

It doesn't look like criterion can produce the system configs but it could be done separately, e.g. after the run and appending to the file if we'd like.

Since I'll add resolc and revive-yul benchmarks posted as PR comments via GHA (in a separate upcoming PR), having those BENCHMARKS.md files tracked by Git may be unnecessary. Thoughts?

crates/resolc/benches/compile.rs
elle-j: Thanks for pointing this out! This path was intended to be updated to an absolute one. Good suggestion btw, I'll add benchmarks for `--yul` and `--standard-json` and use the contracts in the resolc crate.

crates/resolc/benches/compile.rs
elle-j: Will experiment more with this. But when starting with a 10 sample size, the warnings suggested to increase the measurement time to a little under 10 seconds. Using 100 sample size, the warnings suggested to increase the measurement time to ~28 seconds, or reduce the sample size to 10 (e.g. for ERC20).

After replacing the current fixtures used with the ones in resolc, I'll try out more configs.

crates/resolc/Cargo.toml
elle-j: Ah okay, question: As `criterion` is a dev-dependency, and the current `bench-resolc = []` feature does not list other dependencies, does it make it unnecessary to add (and thereby also [this condition](https://github.com/paritytech/revive/blob/b70128902635b59932dcac06fa0b11d8466e3446/crates/resolc/benches/compile.rs#L1))?

crates/resolc/benches/compile.rs
xermicus: I think the right thing to do here is to create multiple benchmarks run with different opt settings.

crates/resolc/benches/compile.rs
xermicus: Compiling the ERC20 to a binary (with full optimizations) takes about 0.5s on a 5year old AMD CPU. So for small cases these seem to be off bit.

crates/resolc/Cargo.toml
xermicus: Sorry I am unsure whether I get the question.

Whether we depend on the `criterion` crate as dev dep or regular dep is irrelevant w.r.t. this feature flag. And yes without the feature being defined the conditional compilation wouldn't work as intended.

Please remove this feature if you are unsure why you added it first hand.

crates/resolc/BENCHMARKS_M4PRO.md
xermicus: I think if they are later going to be checked in by a bot or similar, then they don't need to be checked in with this PR.

crates/resolc/Cargo.toml
elle-j: I originally had these benchmarks in the `revive-benchmarks` crate and used a utility from `resolc`, so I had to add an optional `resolc` dependency to its Cargo.toml, and thereby a `bench-resolc=["resolc"]` feature to only use it for these compilation benchmarks.

I guess when I moved everything over to the `resolc` crate I didn't need that specified in the feature's dependency list anymore, but kept the feature (unnecessarily, as I'm now aware of 🙂)

xermicus: Ah okay please don't blindly copy from `revive-benchmarks`, they require a very specific setup :sweat_smile:

crates/resolc/benches/compile.rs
elle-j: I've [tested out](https://github.com/paritytech/revive/pull/406/commits/4cfaecba078a81354b326d66fa832c636f53dcfa) more combinations of sample size and target measurement time for each group, trying to find a suitable one for each benchmark where the warnings are not emitted.

Essentially aimed for a sample size of 100 for each one if the measurement time could be less than 10 seconds, otherwise I lowered the sample size.

(Made an exception for the `--standard-json` benchmark where multiple contracts exist in the json file, making even the minimum sample size (10) take more than 10 seconds. I could however create another json file containing only 1 contract if we'd like, and benchmark that instead.)

(The criterion table summary is part of the commit I linked to just to easily see the output. Will be removed before merge.)

xermicus: Nice thanks. Yeah the standard json fixture is expected to take a bit longer.

crates/resolc/benches/compile.rs
```diff
@@ -0,0 +1,183 @@
...
+}
+
+fn bench_empty(c: &mut Criterion) {
+    let mut group = group(c, "Empty");
+    group
+        .sample_size(100)
+        .measurement_time(Duration::from_secs(8));
+    let path = absolute_path("src/tests/data/solidity/contract.sol");
+    let resolc_arguments = &[&path, "-O3"];
+    let solc_arguments = &[&path, "--optimize"];
```
xermicus:
```suggestion
    let solc_arguments = &[&path, "--optimize", "--via-ir", "--bin"];
```

I don't think it measures the thing we want without the correct flags.

crates/resolc/benches/compile.rs
```diff
@@ -0,0 +1,183 @@
...
+/// The function under test executes the `solc` executable.
+fn execute_solc(arguments: &[&str], stdin_config: Stdio) {
+    execute_command(
+        SolcCompiler::DEFAULT_EXECUTABLE_NAME,
+        arguments,
+        stdin_config,
+    );
+}
+
+#[inline(always)]
```
xermicus:
```suggestion
```

Is this intentional? We don't blindly sprinkle the code with compiler hints.

xermicus (changes requested):

crates/resolc/benches/compile.rs
xermicus: Can you please add a quick test that the commands we execute actually do compile to binary blobs? The benchmark itself could also assert for the process to return with the success status code.

xermicus: Same for resolc, please add `--bin` (seems like we emit the code anyways, so the question is like why but that's out of scope here).

crates/resolc/benches/compile.rs
elle-j: As this function is called by both functions under test, my intention was to remove this call/return overhead from the measurement by having it be inlined.

crates/resolc/BENCHMARKS_M4PRO.md
kvpanch: > t doesn't look like criterion can produce the system configs but it could be done separately, e.g. after the run and appending to the file if we'd like.

Since all of us have same machine now, it seems ok-ish not to include such information in that file, but maintaining sanity of this file in general is going to be painful if it's unclear which cpu was used for measurements.

As an alternative short-term approach: name this file as `BENCHMARKS_M4PRO.md` while bot that Cyrill mentioned isn't ready. Still not ideal, but helps to track progress without much extra work.
A bit more advanced short-term approach is going to be a combination of [`critcmp`](https://github.com/BurntSushi/critcmp) (measure before/after change) and [`gh pr comment`](https://cli.github.com/manual/gh_pr_comment) to post data to the PR

crates/resolc/benches/compile.rs
xermicus: That's a valid thought :) However consider the impact negligible and you don't know whether the compiler does inline this anyways (and the compiler heuristics are almost always better than what humans think anyways - like you'd add the hint if you found that the compiler doesn't get it right _and_ this has significant impact). So these are almost always unnecessary.

elle-j: Ah okay, thanks :)

crates/resolc/src/tests/data/standard_json/solidity_contracts.json
elle-j, line 1: Changes in this file were to the `settings` field:

```diff
-        "outputSelection": {
-            "*": {
-                "": [
-                    "ast"
-                ]
-            }
-        },
+    "viaIR": true,
+    "outputSelection": {
+      "*": {
+        "": [],
+        "*": ["evm.bytecode"]
+      }
+    },
```

Plus the other parts got reformatted to use 2 spaces.

crates/resolc/BENCHMARKS_M4PRO.md
elle-j: Good idea, I'll rename the file for now.

crates/resolc/benches/utils.rs
```diff
@@ -0,0 +1,264 @@
+use std::{
+    fs::File,
+    path::PathBuf,
+    process::{Command, Output, Stdio},
+};
+
+pub use resolc::{self, SolcCompiler};
```
xermicus: Why is this imported here and then re-exported?

crates/resolc/benches/utils.rs
```diff
@@ -0,0 +1,264 @@
...
+    let absolute_path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join(relative_path);
+    if !absolute_path.is_file() {
+        panic!("expected a file at `{}`", absolute_path.display());
+    }
+
+    absolute_path.to_string_lossy().into_owned()
+}
+
+#[cfg(test)]
+mod tests {
```
xermicus: This needs an `[allow(dead_code)]`.

However: The proper fix seems to implement these helpers in the regular tests. There is even already a `fn execute_command` function. You can just re-use that (DRY principle - maybe move it to `test_utils` so it's usable from within the benchmark). You should probably implement a test there which asserts that the exact command that will be used in the benchmark does produce a binary.

elle-j: You mean moving these test helpers (i.e. `assert_pvm_blob()` etc.) to our `test_utils.rs`? I didn't use the previously existing `execute_command()` since I didn't want to include some overhead that is used in that function. But I'll go ahead and see if it has any meaningful effect on the benchmarks.

crates/resolc/benches/utils.rs
elle-j: `compile.rs` is importing everything that's exported from here (`use utils::*;`), but I could skip the reexport for this line and just explicitly import this in `compile.rs` if that's more idiomatic.

Update: After the refactoring of the CLI utils, the above line is now outdated.

crates/resolc/benches/utils.rs
xermicus: The overhead of a `println!` is going to be insignificant. But can just remove it anyways it's not needed (or replace with a `log::trace!`).

elle-j: Was mainly thinking of the creation of the `stdin_config` and the `CommandResult` fields. Would be nice to export and use just one `execute_command()` tho, so I will check the impact. Can make [these](https://github.com/paritytech/revive/blob/79aaca1a858ac5e82b12413dd1efa85b9e4b0ba9/crates/resolc/src/tests/cli/utils.rs#L99-L100) into functions as well, to only run when explicitly invoked.

elle-j: And seems like it could make more sense to move everything from the benchmark `utils.rs` to the `tests` module (all the helpers, and all the tests).

Perhaps it could also be beneficial to move everything from `src/tests/cli/utils.rs` to `src/test_utils.rs`, in order to keep all test utils together. (The latter could be done as a follow-up PR, and just move the `execute_command()` for now.)

xermicus: > Was mainly thinking of the creation of the stdin_config and the CommandResult fields. Would be nice to export and use just one execute_command() tho, so I will check the impact. 

Not sure what you mean exactly here. I mean the output needs to go somewhere? I doubt that using `-o` to write it into a file and then reading that back is any faster than just grabbing it from STDOUT (probably going via a file is even slower). We are running the full compilation pipeline via the CLI, this includes creating a `resolc` process which then spawns a`solc`, generates code via a recursive process call and then invokes LLVM, LLD, and the PVM linker. Appreciate that there's thoughts behind but please no more bikeshedding of these helpers.

> Perhaps it could also be beneficial to move everything from src/tests/cli/utils.rs to src/test_utils.rs,

Yes if this is now shared code between tests and something else (i.e. benchmarks), it makes sense. I suggest to add a `src/cli_utils.rs` and move them in there.

crates/resolc/src/tests/cli/asm.rs
elle-j, line 1: The only changes made in these preexisting files under `src/tests/cli/*` are updating the imports from `crate::tests::cli::utils` to `crate::cli_utils`.

The only addition is the file `src/tests/cli/bin.rs`.

crates/resolc/benches/compile.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    group.finish();
+}
+
+fn bench_empty(c: &mut Criterion) {
+    let mut group = group(c, "Empty");
+    group
+        .sample_size(100)
+        .measurement_time(Duration::from_secs(8));
+    let path = absolute_path("src/tests/data/solidity/contract.sol");
```
xermicus: Nit: I see those already exists as constants. Can we use the constants here too? Makes it a git more maintainable.

crates/resolc/benches/compile.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    bench(group, resolc_arguments, solc_arguments, None);
+}
+
+fn bench_dependency(c: &mut Criterion) {
+    let mut group = group(c, "Dependency");
+    group
+        .sample_size(50)
+        .measurement_time(Duration::from_secs(9));
+    let path = absolute_path("src/tests/data/solidity/dependency.sol");
```
xermicus: Same for all these following

crates/resolc/src/cli_utils.rs
```diff
@@ -29,15 +30,31 @@ pub const STANDARD_JSON_NO_EVM_CODEGEN_PATH: &str =
...
 /// The `resolc` YUL mode flag.
 pub const RESOLC_YUL_FLAG: &str = "--yul";
 /// The `--yul` option was deprecated in Solidity 0.8.27 in favor of `--strict-assembly`.
-/// See section `--strict-assembly vs. --yul` in https://soliditylang.org/blog/2024/09/04/solidity-0.8.27-release-announcement/
+/// See section `--strict-assembly vs. --yul` in the [release announcement](https://soliditylang.org/blog/2024/09/04/solidity-0.8.27-release-announcement/).
 pub const SOLC_YUL_FLAG: &str = "--strict-assembly";
 
+/// Common `resolc` CLI optimization settings.
+pub struct ResolcOptSettings;
+impl ResolcOptSettings {
```
xermicus, line 46:
```suggestion
pub struct ResolcOptSettings;

impl ResolcOptSettings {
```

crates/resolc/src/cli_utils.rs
```diff
@@ -29,15 +30,31 @@ pub const STANDARD_JSON_NO_EVM_CODEGEN_PATH: &str =
...
+    pub const NONE: &'static str = "-O0";
+    pub const PERFORMANCE: &'static str = "-O3";
+    pub const SIZE: &'static str = "-Oz";
+}
+
+/// Common `solc` CLI optimization settings for `--optimize-runs`.
+pub struct SolcOptSettings;
+impl SolcOptSettings {
```
xermicus, line 55:
```suggestion
pub struct SolcOptSettings;

impl SolcOptSettings {
```

crates/resolc/src/cli_utils.rs
```diff
@@ -29,15 +30,31 @@ pub const STANDARD_JSON_NO_EVM_CODEGEN_PATH: &str =
...
 
 /// The `resolc` YUL mode flag.
 pub const RESOLC_YUL_FLAG: &str = "--yul";
 /// The `--yul` option was deprecated in Solidity 0.8.27 in favor of `--strict-assembly`.
-/// See section `--strict-assembly vs. --yul` in https://soliditylang.org/blog/2024/09/04/solidity-0.8.27-release-announcement/
+/// See section `--strict-assembly vs. --yul` in the [release announcement](https://soliditylang.org/blog/2024/09/04/solidity-0.8.27-release-announcement/).
 pub const SOLC_YUL_FLAG: &str = "--strict-assembly";
 
+/// Common `resolc` CLI optimization settings.
+pub struct ResolcOptSettings;
```
xermicus, line 44: Ah this is a nice idea!

crates/resolc/benches/compile.rs
elle-j: For sure, good call!

Issue #404, opened 2025-11-06
Benchmark Yul to LLVM IR compilation time


elle-j:
Add compilation time benchmarks for:
* Parsing of Yul source code -> AST Object
* AST Object -> LLVM IR

PR #407, opened 2025-11-07
Implement Yul to LLVM IR compilation benchmarks

elle-j:
# Description

Closes [#404](https://github.com/paritytech/revive/issues/404)

Adds compilation time benchmarks for:

* Parsing of Yul source code -> AST Object
* Lowering of AST Object -> LLVM IR (unoptimized)

The benchmarks can be run from the root via:

```sh
# Run all benchmarks in the revive-yul crate (parsing + lowering)
make bench-yul
```

HTML reports will be generated under `target/criterion`, and a summary of the results at [crates/yul/BENCHMARKS_PARSE_M4PRO.md](https://github.com/paritytech/revive/blob/lj/compilation-benchmarks-yul/crates/yul/BENCHMARKS_PARSE_M4PRO.md) and [crates/yul/BENCHMARKS_LOWER_M4PRO.md](https://github.com/paritytech/revive/blob/lj/compilation-benchmarks-yul/crates/yul/BENCHMARKS_LOWER_M4PRO.md) (currently from running on a Mac M4 Pro).

### DIFF
--- a/crates/resolc/src/test_utils.rs
+++ b/crates/resolc/src/test_utils.rs
@@ -486,11 +486,16 @@ pub fn compile_to_yul(
 ) -> String {
     check_dependencies();
 
+    let optimizer = SolcStandardJsonInputSettingsOptimizer::new(
+        solc_optimizer_enabled,
+        Default::default(),
+        Default::default(),
+    );
     let id = CachedBlob {
         contract_name: contract_name.to_owned(),
         solc_optimizer_enabled,
         solidity: source_code.to_owned(),
-        opt: String::new(),
+        opt: optimizer.mode.into(),
     };
 
     if let Some(yul) = YUL_IR_CACHE.lock().unwrap().get(&id) {
@@ -508,11 +513,7 @@ pub fn compile_to_yul(
         Default::default(),
         Default::default(),
         SolcStandardJsonInputSettingsSelection::new_required_for_tests(),
-        SolcStandardJsonInputSettingsOptimizer::new(
-            solc_optimizer_enabled,
-            Default::default(),
-            Default::default(),
-        ),
+        optimizer,
         Default::default(),
         Default::default(),
         Default::default(),


crates/resolc/src/test_utils.rs
```diff
@@ -23,23 +23,27 @@ use revive_solc_json_interface::SolcStandardJsonInputSettingsOptimizer;
...
+static YUL_IR_CACHE: Lazy<Mutex<HashMap<CachedCompilation, String>>> = Lazy::new(Default::default);
 
 const DEBUG_CONFIG: revive_llvm_context::DebugConfig = DebugConfig::new(None, true);
 
 /// Tests may share and re-use contract code.
-/// The compiled blob cache helps avoiding duplicate compilation.
+/// The compilation result (such as blobs) cache helps avoiding duplicate compilation.
 #[derive(Hash, PartialEq, Eq)]
-struct CachedBlob {
+struct CachedCompilation {
```
xermicus: Revert please

crates/resolc/src/test_utils.rs
```diff
@@ -465,3 +469,76 @@ fn compile_evm(
 
     blob
 }
+
+/// Generates Yul IR contracts grouped by files from Solidity source files.
```
xermicus:
```suggestion
/// Internal helper to compiler `sources` to their optimized Yul IR.
```

crates/yul/Cargo.toml
```diff
@@ -18,3 +18,20 @@ thiserror = { workspace = true }
...
 revive-llvm-context = { workspace = true }
+
+[dev-dependencies]
+alloy-primitives = { workspace = true }
+criterion = { workspace = true }
+revive-integration = { workspace = true }
+
+[features]
+bench-parse = []
+bench-llvm-ir = []
```
xermicus: Why are these required?

crates/yul/benches/make_llvm_ir.rs
```diff
@@ -0,0 +1,127 @@
...
+use revive_integration::cases::Contract;
+use revive_llvm_context::{
+    initialize_llvm, Optimizer, OptimizerSettings, PolkaVMContext, PolkaVMTarget, PolkaVMWriteLLVM,
+};
+use revive_yul::{lexer::Lexer, parser::statement::object::Object as AstObject};
+
+/// Function under test - Lowers the Yul AST Object into LLVM IR.
+fn make_llvm_ir(mut ast: AstObject, mut llvm_context: PolkaVMContext) {
+    ast.declare(&mut llvm_context)
+        .expect("expected LLVM IR generation");
```
xermicus:
```suggestion
        .expect("the AST should be valid");
```

Or just `.unwrap()`

crates/yul/benches/make_llvm_ir.rs
```diff
@@ -0,0 +1,127 @@
...
+    ast.into_llvm(&mut llvm_context)
+        .expect("expected LLVM IR generation");
```
xermicus:
```suggestion
        .expect("the AST should lower to LLVM IR");
```

Same here, please either a meaningful message or no message.

crates/yul/benches/make_llvm_ir.rs
```diff
@@ -0,0 +1,127 @@
...
+    BatchSize, BenchmarkGroup, Criterion,
+};
+use inkwell::context::Context as InkwellContext;
+use revive_integration::cases::Contract;
+use revive_llvm_context::{
+    initialize_llvm, Optimizer, OptimizerSettings, PolkaVMContext, PolkaVMTarget, PolkaVMWriteLLVM,
+};
+use revive_yul::{lexer::Lexer, parser::statement::object::Object as AstObject};
+
+/// Function under test - Lowers the Yul AST Object into LLVM IR.
```
xermicus:
```suggestion
/// The function under test lowers the Yul `Object` into LLVM IR.
```

crates/yul/benches/lower.rs
```diff
@@ -0,0 +1,127 @@
...
+    criterion_group, criterion_main,
+    measurement::{Measurement, WallTime},
+    BatchSize, BenchmarkGroup, Criterion,
+};
+use inkwell::context::Context as InkwellContext;
+use revive_integration::cases::Contract;
+use revive_llvm_context::{
+    initialize_llvm, Optimizer, OptimizerSettings, PolkaVMContext, PolkaVMTarget, PolkaVMWriteLLVM,
+};
+use revive_yul::{lexer::Lexer, parser::statement::object::Object as AstObject};
```
xermicus: Why is `Object` aliased?

crates/yul/benches/parse.rs
```diff
@@ -0,0 +1,77 @@
+#![cfg(feature = "bench-parse")]
+
+use alloy_primitives::*;
+use criterion::{
+    criterion_group, criterion_main,
+    measurement::{Measurement, WallTime},
+    BatchSize, BenchmarkGroup, Criterion,
+};
+use revive_integration::cases::Contract;
+use revive_yul::{lexer::Lexer, parser::statement::object::Object as AstObject};
```
xermicus: Why is `Object` aliased?

crates/yul/benches/lower.rs
xermicus, line 1: Can we rename this to something meaningful (e.g. `codegen` or `lowering`)?

crates/yul/benches/make_llvm_ir.rs
```diff
@@ -0,0 +1,127 @@
...
+    mut group: BenchmarkGroup<'_, WallTime>,
+    contract: F,
+    optimizer_settings: OptimizerSettings,
+) where
+    F: Fn() -> Contract,
+{
+    let llvm = InkwellContext::create();
+    let ast = parse(&contract().yul);
+
+    group.sample_size(10);
```
xermicus: Why 10? Seems too low.

crates/yul/benches/make_llvm_ir.rs
```diff
@@ -0,0 +1,127 @@
...
+};
+use inkwell::context::Context as InkwellContext;
+use revive_integration::cases::Contract;
+use revive_llvm_context::{
+    initialize_llvm, Optimizer, OptimizerSettings, PolkaVMContext, PolkaVMTarget, PolkaVMWriteLLVM,
+};
+use revive_yul::{lexer::Lexer, parser::statement::object::Object as AstObject};
+
+/// Function under test - Lowers the Yul AST Object into LLVM IR.
+fn make_llvm_ir(mut ast: AstObject, mut llvm_context: PolkaVMContext) {
```
xermicus: Please rename to something more meaningful

crates/yul/benches/parse.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Function under test - Parses the Yul `source_code`.
```
xermicus:
```suggestion
/// The function under test parses the Yul `source_code`.
```

crates/yul/benches/parse.rs
```diff
@@ -0,0 +1,77 @@
...
+/// Consumes the source code String in order to not time the cloning.
```
xermicus:
```suggestion
```

The comment is confusing (you pass it by value so the comment doesn't add semantic information - also this is just a benchmarking harness technicality and has nothing to do with the parser but the comment indicates otherwise).

crates/yul/benches/parse.rs
```diff
@@ -0,0 +1,77 @@
...
+
+fn bench<F>(mut group: BenchmarkGroup<'_, WallTime>, contract: F)
+where
+    F: Fn() -> Contract,
+{
+    let source_code = contract().yul;
+
+    group.sample_size(10);
+
+    group.bench_function("Revive", |b| {
```
xermicus:
```suggestion
    group.bench_function("parse", |b| {
```

crates/yul/benches/parse.rs
```diff
@@ -0,0 +1,77 @@
...
+        b.iter_batched(|| source_code.clone(), parse, BatchSize::SmallInput);
+    });
+
+    group.finish();
+}
+
+fn bench_baseline(c: &mut Criterion) {
+    bench(group(c, "Baseline - Parse"), Contract::baseline);
```
xermicus:
```suggestion
    bench(group(c, "Baseline"), Contract::baseline);
```

Same for all of them

crates/yul/benches/make_llvm_ir.rs
```diff
@@ -0,0 +1,127 @@
...
+    optimizer_settings: OptimizerSettings,
+) where
+    F: Fn() -> Contract,
+{
+    let llvm = InkwellContext::create();
+    let ast = parse(&contract().yul);
+
+    group.sample_size(10);
+
+    group.bench_function("Revive", |b| {
```
xermicus:
```suggestion
    group.bench_function("codegen", |b| {
```

Or whatever we want to call the benchmark

crates/yul/benches/make_llvm_ir.rs
```diff
@@ -0,0 +1,127 @@
...
+            BatchSize::SmallInput,
+        );
+    });
+
+    group.finish();
+}
+
+fn bench_baseline(c: &mut Criterion) {
+    bench(
+        group(c, "Baseline - To LLVM IR"),
```
xermicus:
```suggestion
        group(c, "Baseline"),
```

Same for all of them

crates/yul/benches/lower.rs
```diff
@@ -0,0 +1,127 @@
...
+    F: Fn() -> Contract,
+{
+    let llvm = InkwellContext::create();
+    let ast = parse(&contract().yul);
+
+    group.sample_size(10);
+
+    group.bench_function("Revive", |b| {
+        b.iter_batched(
+            || (ast.clone(), create_llvm_context(&llvm, &optimizer_settings)),
```
xermicus: The llvm context should probably created here instead. [It's not thread safe.](https://thedan64.github.io/inkwell/inkwell/context/struct.Context.html)

xermicus (changes requested):
Quick first round

crates/yul/Cargo.toml
elle-j: As with the other PR, I've removed the unnecessary flags 👍

crates/yul/benches/make_llvm_ir.rs
elle-j: I've updated to make it about as big as I could within the default time 👍

crates/yul/benches/lower.rs
elle-j: Thought it'd add a bit more meaning when used by itself in the code without its path.

crates/yul/benches/parse.rs
elle-j: Yeah I could have clarified a bit more. Essentially what I was referring to was that by passing `String` (and not a reference) and taking ownership of it, the function can call `Lexer::new(source_code)` instead of `Lexer::new(source_code.to_owned()`, thereby not timing the copying of the string (`to_owned()`). Can remove it tho ✅

crates/yul/benches/make_llvm_ir.rs
elle-j: See [previous comment](https://github.com/paritytech/revive/pull/407#discussion_r2542361859) 👍

crates/yul/benches/lower.rs
elle-j: From my understanding, a [single thread](https://doc.rust-lang.org/cargo/commands/cargo-bench.html#miscellaneous-options) is used:

> The Rust test harness runs benchmarks serially in a single thread.

In the current implementation, `|| (ast.clone(), create_llvm_context(&llvm, &optimizer_settings)),` is the per-iteration setup that is not timed.

crates/yul/benches/parse.rs
elle-j: This had to do with criterion benchmark groups and how criterion-table's tables are produced.

The `BenchmarkGroup` name needs to be unique among all the benchmark groups run. So if having e.g. the name `ERC20` in both `parse.rs` and `lower.rs` and running them together (`make bench-yul`), we'll get the warning:

> Benchmark group ERC20 encountered again. Benchmark group IDs must be unique.

It looks like this could mess with the reporting and how the data is aggregated.

I've [refactored](https://github.com/paritytech/revive/blob/05c37b412b05a9042d3f0d1108efdfabc4087b18/crates/yul/benches/parse.rs#L20) this suffix tho, into one place (the `group()` function).

crates/llvm-context/src/polkavm/context/test_utils.rs
xermicus, line 1: Please remove this module.

The exported function is unnecessary. If anything, this should be another shortcut constructor on `Context` (e.g. `new_dummy`).

crates/resolc/src/test_utils.rs
```diff
@@ -475,3 +477,103 @@ fn compile_evm(
 
     blob
 }
+
+/// Internal helper to compiler `sources` to their optimized Yul IR.
+fn generate_yul(
```
xermicus: Inline into `compile_to_yul`

crates/resolc/src/test_utils.rs
```diff
@@ -475,3 +477,103 @@ fn compile_evm(
...
+    Ok(output.contracts)
+}
+
+/// Compiles the Solidity source code into Yul IR and returns
+/// the Yul IR code of the contract named `contract_name`.
+/// The `solc` optimizer will be enabled.
+pub fn compile_to_yul(contract_name: &str, source_code: &str) -> String {
+    compile_to_yul_with_options(contract_name, source_code, true)
+}
+
```
xermicus:
```suggestion
```

Dead code

crates/resolc/src/test_utils.rs
```diff
@@ -475,3 +477,103 @@ fn compile_evm(
...
+/// Compiles the Solidity source code into Yul IR and returns
+/// the Yul IR code of the contract named `contract_name`.
+pub fn compile_to_yul_with_options(
```
xermicus:
```suggestion
pub fn compile_to_yul(
```

This is a new API and we don't need to preserve the old pattern.

crates/yul/benches/lower.rs
xermicus: I always prefer to use the parent path instead of aliasing when there's a clash. But here I don't even see a naming clash with another `Object` symbol? Please remove the alias for some basic consistency with the codebase.

Makefile
```diff
@@ -69,6 +71,9 @@ test-integration: install-bin
 test-resolc: install
 	cargo test --package resolc --benches
 
+test-yul:
+	cargo test --package revive-yul --benches
+
```
xermicus, line 76: I think `test-yul` should be added to the `test` target as well.

crates/yul/benches/parse.rs
```diff
@@ -0,0 +1,72 @@
...
+    measurement::{Measurement, WallTime},
+    BenchmarkGroup, Criterion,
+};
+use revive_integration::cases::Contract;
+use revive_yul::{lexer::Lexer, parser::statement::object::Object as AstObject};
+
+/// The function under test parses the Yul `source_code`.
+fn parse(source_code: &str) {
+    let mut lexer = Lexer::new(source_code.to_owned());
+    AstObject::parse(&mut lexer, None).expect("expected a Yul AST Object");
```
xermicus:
```suggestion
    AstObject::parse(&mut lexer, None).expect("the Yul source should parse");
```

Please provide either no message (`unwrap()`) or a meaningful message.

crates/yul/benches/lower.rs
```diff
@@ -0,0 +1,101 @@
...
+fn lower(mut ast: AstObject, mut llvm_context: PolkaVMContext) {
+    ast.declare(&mut llvm_context)
+        .expect("the AST should be valid");
+    ast.into_llvm(&mut llvm_context)
+        .expect("the AST should lower to LLVM IR");
+}
+
+fn parse(source_code: &str) -> AstObject {
+    let mut lexer = Lexer::new(source_code.to_owned());
+    AstObject::parse(&mut lexer, None).expect("expected a Yul AST Object")
```
xermicus:
```suggestion
    AstObject::parse(&mut lexer, None).expect("the Yul source should parse")
```

crates/yul/benches/lower.rs
```diff
@@ -0,0 +1,101 @@
...
+
+fn bench<F>(mut group: BenchmarkGroup<'_, WallTime>, contract: F)
+where
+    F: Fn() -> Contract,
+{
+    let ast = parse(&contract().yul);
+    let llvm = InkwellContext::create();
+    // The optimizer settings will not affect the benchmarks since we're currently
+    // measuring lowering into unoptimized IR and not running the optimization passes.
+    // (This is still needed as an argument below.)
```
xermicus:
```suggestion
```

crates/yul/benches/lower.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+fn bench<F>(mut group: BenchmarkGroup<'_, WallTime>, contract: F)
+where
+    F: Fn() -> Contract,
+{
+    let ast = parse(&contract().yul);
+    let llvm = InkwellContext::create();
+    // The optimizer settings will not affect the benchmarks since we're currently
+    // measuring lowering into unoptimized IR and not running the optimization passes.
```
xermicus:
```suggestion
    // not running the optimization passes.
```

crates/yul/benches/lower.rs
```diff
@@ -0,0 +1,101 @@
...
+    c.benchmark_group(format!("{group_name} - lower"))
+}
+
+fn bench<F>(mut group: BenchmarkGroup<'_, WallTime>, contract: F)
+where
+    F: Fn() -> Contract,
+{
+    let ast = parse(&contract().yul);
+    let llvm = InkwellContext::create();
+    // The optimizer settings will not affect the benchmarks since we're currently
```
xermicus:
```suggestion
    // The optimizer settings will not affect the benchmarks since we're
```

crates/yul/benches/parse.rs
xermicus: I assumed we'll have two tables as these measure two things. Doing it like this seems more like a hack. But I guess it works fine.

Makefile
elle-j: These tests are included via `test-workspace` (ran by `make test`). Although, I've now [added](https://github.com/paritytech/revive/pull/407/commits/e8d5d58d68843e5007b9ba791f1e363f601afc4c) `--benches` to the command run via `test-workspace`, so all bench targets are tested.

xermicus: I like the idea. We can drive this even further: Instead of add `--benches` to that target, we can just add `--all-targets` there. WDYT?

crates/llvm-context/src/polkavm/context/test_utils.rs
xermicus: Just realized it wasn't entirely clear. I meant that this whole helper function can be boiled down to:

```Rust
    Context::new(
        llvm,
        llvm.create_module("dummy"),
        Optimizer::new(optimizer_settings),
        Default::default(),
        Default::default(),
    )
```

But it's bikeshedding at that point. While a `new_dummy` constructor that just folds into above code seems easier, I'll leave it up to you.

crates/yul/benches/parse.rs
elle-j: Yeah, that was one way of getting around this uniqueness issue, and preventing the results from being combined into one table. I've [separated](https://github.com/paritytech/revive/pull/407/commits/87c1117a85370fc87d48341c8b255ca2c2421547) out the results into separate files now instead.

crates/llvm-context/src/polkavm/context/test_utils.rs
elle-j: Added `new_dummy` 👍

Makefile
elle-j: For sure 🙂 ✅

Makefile
```diff
@@ -14,13 +14,17 @@
...
 	test-workspace \
 	test-wasm \
 	test-llvm-builder
 	bench \
 	bench-pvm \
 	bench-evm \
 	bench-resolc \
+	bench-yul \
+	bench-parse-yul \
+	bench-lower-yul \
```
xermicus:
```suggestion
```

Makefile
```diff
@@ -95,6 +102,16 @@ bench-resolc: test-resolc
...
 
+bench-yul: test-yul bench-parse-yul bench-lower-yul
+
+bench-parse-yul: test-yul
+	cargo criterion --package revive-yul --bench parse --message-format=json \
+	| criterion-table > crates/yul/BENCHMARKS_PARSE_M4PRO.md
+
+bench-lower-yul: test-yul
+	cargo criterion --package revive-yul --bench lower --message-format=json \
+	| criterion-table > crates/yul/BENCHMARKS_LOWER_M4PRO.md
```
xermicus:
```suggestion
bench-yul: test-yul
	cargo criterion --package revive-yul --bench parse --message-format=json \
	| criterion-table > crates/yul/BENCHMARKS_PARSE_M4PRO.md
	cargo criterion --package revive-yul --bench lower --message-format=json \
	| criterion-table > crates/yul/BENCHMARKS_LOWER_M4PRO.md
```

Let's not make it too fine granular.

crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -256,6 +256,20 @@ impl<'ctx> Context<'ctx> {
         }
     }
 
+    /// Initializes a new dummy LLVM context.
```
xermicus, line 259:
```suggestion
    /// Initializes a new dummy LLVM context.
    ///
    /// Omits the LLVM module initialization; use this only in tests and benchmarks.
```

crates/resolc/src/test_utils.rs
```diff
@@ -33,6 +33,7 @@ static PVM_BLOB_CACHE: Lazy<Mutex<HashMap<CachedBlob, Vec<u8>>>> = Lazy::new(Def
 static EVM_BLOB_CACHE: Lazy<Mutex<HashMap<CachedBlob, Vec<u8>>>> = Lazy::new(Default::default);
 static EVM_RUNTIME_BLOB_CACHE: Lazy<Mutex<HashMap<CachedBlob, Vec<u8>>>> =
     Lazy::new(Default::default);
+static YUL_IR_CACHE: Lazy<Mutex<HashMap<CachedBlob, String>>> = Lazy::new(Default::default);
```
xermicus, line 36: Dead code; revert this or (preferred) make use of the cache.

crates/resolc/src/test_utils.rs
```diff
@@ -475,3 +476,90 @@ fn compile_evm(
...
+        contract_name: contract_name.to_owned(),
+        solc_optimizer_enabled,
+        solidity: source_code.to_owned(),
+        opt: String::new(),
+    };
+
+    if let Some(yul) = YUL_IR_CACHE.lock().unwrap().get(&id) {
+        return yul.clone();
+    }
+
```
xermicus, line 504: Dead code because you never populate it.

xermicus (commented):
Getting there

crates/resolc/src/test_utils.rs
elle-j: Missed the insert, thanks for catching this

crates/resolc/src/test_utils.rs
```diff
@@ -475,3 +476,94 @@ fn compile_evm(
...
+    solc_optimizer_enabled: bool,
+) -> String {
+    check_dependencies();
+
+    let id = CachedBlob {
+        contract_name: contract_name.to_owned(),
+        solc_optimizer_enabled,
+        solidity: source_code.to_owned(),
+        opt: String::new(),
+    };
```
xermicus, line 499:
```suggestion
    let optimizer = SolcStandardJsonInputSettingsOptimizer::new(
        solc_optimizer_enabled,
        Default::default(),
        Default::default(),
    );
    let id = CachedBlob {
        contract_name: contract_name.to_owned(),
        solc_optimizer_enabled,
        solidity: source_code.to_owned(),
        opt: optimizer.mode.into(),
    };
```

The empty optimizer is only valid when compiling EVM bytecode. This fixes the ID.

crates/resolc/src/test_utils.rs
```diff
@@ -475,3 +476,94 @@ fn compile_evm(
...
+        None,
+        sources.clone(),
+        Default::default(),
+        Default::default(),
+        SolcStandardJsonInputSettingsSelection::new_required_for_tests(),
+        SolcStandardJsonInputSettingsOptimizer::new(
+            solc_optimizer_enabled,
+            Default::default(),
+            Default::default(),
+        ),
```
xermicus:
```suggestion
        optimizer,
```

Let's also make it explicit. This removes room for nasty bugs in the future (for example when we change the default away from size opts).

crates/resolc/src/test_utils.rs
elle-j: Ah okay 👍 Regarding your [other comment](https://github.com/paritytech/revive/pull/407#discussion_r2558720309) about making it explicit, so rather than the above suggestion, make the `mode` explicit like so?:

```diff
    let optimizer = SolcStandardJsonInputSettingsOptimizer::new(
        solc_optimizer_enabled,
-       Default::default(),
+       OptimizerSettings::size()
+           .middle_end_as_string()
+           .chars()
+           .last()
+           .unwrap(),
        Default::default(),
    );
```

xermicus: No you should just be able to merge my two suggestions in batch. What's in the two suggestions is what I meant with making it explicit (like we explicitly instantiate the optimizer and use that - maybe not the best wording).

PR #410, opened 2025-11-12
std json: do not request EVM bytecode

xermicus:
These compiler outputs are irrelevant for the `resolc` compilation pipeline. It can also break the compilation pipeline. For example, solc may reject the code without via-ir or fail during the EVM codegen phase with stack too deep errors - which is not relevant to us.


0xOmarA (approved):
Looks good to me, thanks for the quick fix!

PR #418, opened 2025-11-21
Fetch LLVM release according to branch in llvm submodule

kvpanch:
This change is essential first step to unblock gracefully enable LLVM 21 in revive compiler as current action blindly fetched latest LLVM binary from the list of releases on our page.
The change queries release version from the submodule and uses it. Next step: build LLVM 21 release binary
Next next step: switch revive compile to LLVM 21 after all issues are fixed

### DIFF
--- a/.github/actions/get-llvm/action.yml
+++ b/.github/actions/get-llvm/action.yml
@@ -7,23 +7,56 @@ name: "Download LLVM"
 inputs:
   target:
     required: true
+  version:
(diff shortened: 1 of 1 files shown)


Cargo.lock
xermicus, line 1: I think this got out of date with some past PR.

I just updated the dependencies and it should be up to date again. Can you please update this branch with `main` and remove the `Cargo.lock` in this PR (it should no longer diff anyways)?

kvpanch: absolutely

xermicus: thanks

PR #422, opened 2025-11-26
ci: Add GHA workflow executing benchmarks

### LINKED ISSUE #405: Add GHA workflow executing benchmarks
Add a workflow for executing the resolc and yul benchmarks after all other jobs have finished. Post the diff against `main` as a PR comment.

elle-j:
## Description

Adds a workflow executing the resolc and yul benchmarks.

* The workflow can be manually triggered via the Actions UI.
* User input:
  * Branch, tag, or SHA before changes (e.g. `main`) _(required)_
  * Branch, tag, or SHA after changes (e.g. `lj/ci-benchmarks`) _(required)_
  * PR number  _(optional)_
    * If provided, the benchmarking result will be posted as a PR comment
      * The comment is updated if it exists, otherwise it is created
    * If omitted, the result is uploaded as an artifact

This PR originally implemented the workflow to automatically run on each PR once its tests passed successfully. As we likely do not need to run this workflow for each PR, the workflow has been rewritten to manually trigger it.

Providing user input also allows for comparisons between any commits, such as releases.

### File must exist on the default branch

* This workflow is triggered by a `workflow_dispatch` event [only](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch) if the workflow file exists on the default branch. **Therefore, the workflow cannot be run for this PR.** (See section below on how it has still been tested.)
* Once merged, for any subsequent PRs that modify this workflow, those changes will be testable before merge (by choosing to run the workflow from the PR branch).

## Testing the Workflow

I've tested a similar enough workflow in a personal repo in order to push it to `main` and trigger it. Will verify again once this PR is merged.

- [x] Runs when provided with branch names
- [x] Runs when provided with SHAs
- [x] A PR comment is created if a PR number is provided and the comment does not exist
- [x] A PR comment is updated if a PR number is provided and the comment already exists
- [x] An artifact is created if a PR number is omitted
- [x] An in-progress run is cancelled if another one is triggered comparing the same refs

## Example PR Comment

<details>
    <summary>Expand to see an example PR comment</summary>

# Benchmarks

* **Ref Before:** `main`
* **Ref After:** `lj/ci-benchmarks`

## Compile E2E

<details>
    <summary>Results</summary>

```
group                                            resolc_before                            resolc_changes
-----                                            -------------                            --------------
Dependency/resolc                                1.00    144.0±6.94ms        ? ?/sec      1.00    144.6±5.97ms        ? ?/sec
Dependency/solc                                  1.04    65.9±10.73ms        ? ?/sec      1.00    63.5±10.10ms        ? ?/sec
Empty/resolc                                     1.03     68.8±4.40ms        ? ?/sec      1.00     66.6±4.61ms        ? ?/sec
Empty/solc                                       1.13     11.2±2.01ms        ? ?/sec      1.00     10.0±2.28ms        ? ?/sec
LargeDivRem/resolc                               1.00    119.3±5.62ms        ? ?/sec      1.02   121.3±17.03ms        ? ?/sec
LargeDivRem/solc                                 1.15     24.7±3.15ms        ? ?/sec      1.00     21.5±1.03ms        ? ?/sec
Memset (`--yul`)/resolc                          1.02     62.7±3.92ms        ? ?/sec      1.00     61.7±3.76ms        ? ?/sec
Memset (`--yul`)/solc                            1.00      9.0±0.84ms        ? ?/sec      1.12     10.1±2.23ms        ? ?/sec
Multiple Contracts (`--standard-json`)/resolc    1.01  1559.2±68.95ms        ? ?/sec      1.00  1539.3±64.92ms        ? ?/sec
Multiple Contracts (`--standard-json`)/solc      1.01   652.0±19.38ms        ? ?/sec      1.00   648.4±31.83ms        ? ?/sec
Return (`--yul`)/resolc                          1.00    60.1±17.02ms        ? ?/sec      1.02    61.2±14.50ms        ? ?/sec
Return (`--yul`)/solc                            1.09      8.8±2.37ms        ? ?/sec      1.00      8.0±3.08ms        ? ?/sec
```

</details>

## Parse Yul

<details>
    <summary>Results</summary>

```
group             yul_parse_before                         yul_parse_changes
-----             ----------------                         -----------------
Baseline/parse    1.07      8.9±3.64µs        ? ?/sec      1.00      8.4±0.09µs        ? ?/sec
ERC20/parse       1.00   162.5±34.25µs        ? ?/sec      1.07  173.3±123.25µs        ? ?/sec
SHA1/parse        1.01    80.8±33.04µs        ? ?/sec      1.00    80.3±52.82µs        ? ?/sec
Storage/parse     1.00     16.9±8.15µs        ? ?/sec      1.01     17.2±7.54µs        ? ?/sec
Transfer/parse    1.00     18.5±0.23µs        ? ?/sec      1.23    22.7±18.50µs        ? ?/sec
```

</details>

## Lower Yul

<details>
    <summary>Results</summary>

```
group             yul_lower_before                         yul_lower_changes
-----             ----------------                         -----------------
Baseline/lower    1.00    122.1±5.95µs        ? ?/sec      1.09   133.6±35.12µs        ? ?/sec
ERC20/lower       1.00   641.7±13.24µs        ? ?/sec      1.15  735.6±665.92µs        ? ?/sec
SHA1/lower        1.00   361.4±78.47µs        ? ?/sec      1.09  392.3±114.48µs        ? ?/sec
Storage/lower     1.00   167.2±50.34µs        ? ?/sec      1.01   168.2±32.45µs        ? ?/sec
Transfer/lower    1.00    167.5±8.70µs        ? ?/sec      1.03   173.1±16.32µs        ? ?/sec
```

</details>

</details>

## Resolved Issues

Closes #405

### DIFF
--- a/.github/scripts/resolc-bin-gh-pages.sh
+++ b/.github/scripts/resolc-bin-gh-pages.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 # This script updates `index.md` in the GitHub Pages root directory provided
-# by the required argument to be passed. The file will be updated to simply
-# render the `resolc-bin` JSON data for each of the supported platforms.
+# by the required argument to be passed. The file will be updated to render
+# the `resolc-bin` release data as a formatted table for each supported platform.
 # `index.md` is the file served by GitHub Pages after being built by Jekyll
 # and the Markdown processed by kramdown.
 
@@ -33,20 +33,22 @@ for file in "${build_info_files[@]}"; do
     fi
 done
 
-# Sort the data by version in descending order.
-sort_by_version_descending() {
+# Render builds as a markdown table with clickable links
+render_builds_table() {
     local build_info_file="$1"
     if [ -z "$build_info_file" ]; then
-        echo "Error: A file path argument is required for sorting"
+        echo "Error: A file path argument is required"
         return 1
     fi
 
-    # Load the data and sort builds and releases with the latest version first.
-    data=$(jq '.' "$build_info_file")
-    sorted_data=$(echo "$data" | jq '.builds = (.builds | sort_by(.version) | reverse) |
-                                     .releases = (.releases | to_entries | sort_by(.key) | reverse | from_entries)')
-
-    echo "$sorted_data" | jq .
+    # Sort builds by version descending and render as markdown table
+    jq -r '
+        .builds | sort_by(.version) | reverse |
+        ["| Release | Solc Versions | SHA256 |", "|---------|---------------|--------|"] +
+        [.[] |
+            "| [\(.name) \(.longVersion)](\(.url)) | \(.firstSolcVersion) - \(.lastSolcVersion) | `\(.sha256[0:16])...` |"
+        ] | .[]
+    ' "$build_info_file"
 }
 
 echo "Updating GitHub Pages index.md file..."
@@ -64,92 +66,92 @@ The information is synced with the [resolc-bin GitHub repository](https://github
 ## Linux
 
 <details>
-    <summary>See builds</summary>
+<summary>See builds</summary>
 
-{% highlight json %}
-$(sort_by_version_descending $linux)
-{% endhighlight %}
+$(render_builds_table $linux)
 
 </details>
+
+[JSON](https://paritytech.github.io/resolc-bin/linux/list.json)
 
 ## MacOS
 
 <details>
-    <summary>See builds</summary>
+<summary>See builds</summary>
 
-{% highlight json %}
-$(sort_by_version_descending $macos)
-{% endhighlight %}
+$(render_builds_table $macos)
 
 </details>
+
+[JSON](https://paritytech.github.io/resolc-bin/macos/list.json)
 
 ## Wasm
 
 <details>
-    <summary>See builds</summary>
+<summary>See builds</summary>
 
-{% highlight json %}
-$(sort_by_version_descending $wasm)
-{% endhighlight %}
+$(render_builds_table $wasm)
 
 </details>
+
+[JSON](https://paritytech.github.io/resolc-bin/wasm/list.json)
 
 ## Windows
 
 <details>
-    <summary>See builds</summary>
+<summary>See builds</summary>
 
-{% highlight json %}
-$(sort_by_version_descending $windows)
-{% endhighlight %}
+$(render_builds_table $windows)
 
 </details>
+
+[JSON](https://paritytech.github.io/resolc-bin/windows/list.json)
 
 ## Nightly
 
 ### Linux
 
 <details>
-    <summary>See builds</summary>
+<summary>See builds</summary>
 
-{% highlight json %}
-$(sort_by_version_descending $nightly_linux)
-{% endhighlight %}
+$(render_builds_table $nightly_linux)
 
 </details>
+
+[JSON](https://paritytech.github.io/resolc-bin/nightly/linux/list.json)
 
 ### MacOS
 
 <details>
-    <summary>See builds</summary>
+<summary>See builds</summary>
 
-{% highlight json %}
-$(sort_by_version_descending $nightly_macos)
-{% endhighlight %}
+$(render_builds_table $nightly_macos)
 
 </details>
+
+[JSON](https://paritytech.github.io/resolc-bin/nightly/macos/list.json)
 
 ### Wasm
 
 <details>
-    <summary>See builds</summary>
+<summary>See builds</summary>
 
-{% highlight json %}
-$(sort_by_version_descending $nightly_wasm)
-{% endhighlight %}
+$(render_builds_table $nightly_wasm)
 
 </details>
+
+[JSON](https://paritytech.github.io/resolc-bin/nightly/wasm/list.json)
 
 ### Windows
 
 <details>
-    <summary>See builds</summary>
+<summary>See builds</summary>
 
-{% highlight json %}
-$(sort_by_version_descending $nightly_windows)
-{% endhighlight %}
+$(render_builds_table $nightly_windows)
 
 </details>
+
+[JSON](https://paritytech.github.io/resolc-bin/nightly/windows/list.json)
 EOF
 
 echo "File has been updated!"


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,119 @@
...
+          git fetch origin main
+          git checkout -f main
+          make install
+
+      - name: Run Benchmarks on Main Branch
+        run: |
+          cargo bench --package resolc --bench compile -- --save-baseline main_resolc
+          cargo bench --package revive-yul --bench parse -- --save-baseline main_yul_parse
+          cargo bench --package revive-yul --bench lower -- --save-baseline main_yul_lower
+        timeout-minutes: 20
```
elle-j, line 67: We could also remove the `timeout-minutes` and add it only if we notice unusually long runs. (This took about 6-7min on M4 Pro.)

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,119 @@
...
+          echo "LLVM_SYS_181_PREFIX=$(pwd)/llvm-x86_64-unknown-linux-gnu" >> $GITHUB_ENV
+
+      - name: Install Benchmarking Tools
+        run: |
+          cargo install critcmp
+
+      - name: Checkout and Compile Main Branch
+        run: |
+          git fetch origin main
+          git checkout -f main
```
kvpanch: IMO, that makes sense to parameterize it by allowing to specify branch or tag to test against. This will help to nicely measure improvements/regressions over some release.
Obviously, that can be done later, not in this PR

kvpanch (commented):
Is it known if CI machine is good for such perf testing ? For example, can CI machine run several tasks in parallel ? If so, perf results might not be stable and it might be required to run such perf testing on a dedicated machine

.github/workflows/benchmark.yml
elle-j: That can be a neat and valuable feature, thanks. Can add in a follow-up 👍 

Regarding your other comment and concurrency, jobs and workflows (not steps within jobs) can run concurrently by default. Which, thanks for the reminder, made me [add](https://github.com/paritytech/revive/pull/422/commits/6187f76d630de762dfb064b969f3cc6dcd6e844b) a `concurrency` group where this workflow [only runs one at a time](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow) for the given PR branch (and cancels in-progress ones if it's triggered again while in progress).

I'd have to look up whether another CI machine is still more optimial for perf testing tho.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,123 @@
...
+    workflows: [Test]
+    types: [completed]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-24.04
```
xermicus: I'd like to use the same host we benchmark the polkadot SDK with.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,123 @@
...
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request' }}
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Checkout PR Branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.workflow_run.head_sha }}
```
xermicus: Why are we checking out the PR branch here only to checkout the `main` branch first and then switch to the PR branch again?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,123 @@
...
+
+name: Benchmark
+
+on:
+  workflow_run:
+    workflows: [Test]
+    types: [completed]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_branch }}
```
xermicus: I'd appreciate a comment here why this is needed and what it does.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,123 @@
...
+          \`\`\`
+          $(cat benchmarks_resolc)
+          \`\`\`
+
+          </details>
+
+          ## Parse Yul
+
+          <details>
+              <summary>Results</summary>
```
xermicus, line 124: Hiding the results is definitively what we want here in order to not make it too noisy. Can we also have a summary or something on top that tells PR authors quickly whether the benchmarks have regressed, improved or no significant change was detected?

.github/workflows/benchmark.yml
elle-j: Seems like different ones are being used for benches in the Polkadot SDK, but two common ones are e.g. [`parity-large`](https://github.com/paritytech/polkadot-sdk/blob/285fb47ff1fe6d53c3c6ac96e18ab2173e05ed9c/.github/workflows/reusable-preflight.yml#L172) and [`parity-benchmark`](https://github.com/paritytech/polkadot-sdk/blob/285fb47ff1fe6d53c3c6ac96e18ab2173e05ed9c/.github/workflows/reusable-preflight.yml#L162) (both enabled in our revive repo). The differences can be [seen here](https://github.com/paritytech/ci_cd/wiki/GitHub#paritytech-self-hosted-runners).

.github/workflows/benchmark.yml
elle-j: Essentially due to the `get-solc` and `get-llvm` actions used first. I figured we'd want to use these action files from the PR branch in case those were modified as part of the PR. An alternative would be to run those steps twice (once on `main` and then on the PR branch after switching).

.github/workflows/benchmark.yml
elle-j: Can add that it cancels in-progress runs (for runs on the same branch as the PR) if the workflow is re-triggered during a current run (by e.g. another push to the PR and successful `Test` workflow run).

.github/workflows/benchmark.yml
xermicus: I think we should narrow down the workflow trigger. Thinking about it, this will be generating a lot of CI hours. I don't think it's necessary to run this workflow in every PR. Is there a simple way to configure it to to be manually triggered by the PR author?

.github/workflows/benchmark.yml
xermicus: I mean either ways if they are modified in an incompatible way either main or the PR branch is broken. I was just thinking it would be simpler to first bench the CI branch and then checking out main for the comparison.

.github/workflows/benchmark.yml
elle-j: I agree, also thought about that once I was off. I've rewritten the workflow to instead be manually triggered via the Actions UI. (See updated PR description.)

.github/workflows/benchmark.yml
elle-j: Update: Regarding parameterization, I've incorporated this into the rewrite of this workflow that I've now done in order to manually trigger this workflow on demand, rather than automatically on each PR which may not be necessary.

.github/workflows/benchmark.yml
elle-j: Update: Opted for `parity-large` in our case.

.github/workflows/benchmark.yml
elle-j: I will make a note of this for a follow-up 👍 Will likely require some parsing of critcmp's output. But def improves the readability.

elle-j:
I've slightly rewritten the workflow to instead be manually triggered, rather than running this on each PR and each of its updates. We can now also compare any branch, SHA, or tag. (See updated PR description.)

PR #423, opened 2025-11-26
Clone emsdk for release-llvm CI action

kvpanch:
Cloning of emsdk has been accidentally removed, that caused inability to
build LLVM release with current main.
The change adds this cloning back before emsdk script is sourced.

### DIFF
--- a/.github/workflows/release-llvm.yml
+++ b/.github/workflows/release-llvm.yml
@@ -104,7 +104,7 @@ jobs:
       - name: Build LLVM
         if: ${{ matrix.target == 'wasm32-unknown-emscripten' }}
         run: |
-          revive-llvm emsdk clone
+          revive-llvm emsdk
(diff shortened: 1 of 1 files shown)


kvpanch:
https://github.com/paritytech/revive/actions/runs/19716663590/job/56490325088

elle-j (approved):

kvpanch:
> ```diff
> crates/llvm-builder/src/revive_llvm/main.rs
> ```

Doing this way will couple "emsdk clone" and build into a single command. Prior to "llvm submodule" change, cloning was done separately with`revive-llvm  --target-env emsdk clone` command, but `clone` was replaces with `emsdk` as only `emsdk` needs to be cloned.
btw, we can also move `emsdk` to submodule too, but keep it uninitialized in most cases.

PR #425, opened 2025-12-02
Update ubuntu's source list to use `nobel`

kvpanch:
This helps to install clang-18 and not clang-15. Latter cannot be used anymore to build new LLVM due to some missed clang option.


kvpanch:
context: while trying to get release build of LLVM 21 I got https://github.com/paritytech/revive/actions/runs/19830461975/job/56814914090 problem, which was due to use of clang-15 during the build, that is default in jammy fish (ubuntu 22.04). That's odd as used ubuntu in CI is 24.04, which is noble.
Replacing sources to get packages helps to get clang-18  that can build LLVM 21

.github/workflows/release-llvm.yml
```diff
@@ -82,7 +80,10 @@ jobs:
       - name: Install Dependencies
         if: ${{ matrix.host == 'linux' }}
         run: |
-          sudo apt-get update && sudo apt-get install -y cmake ninja-build curl git libssl-dev pkg-config clang lld musl
+          cat /etc/apt/sources.list
+          sudo sed -i 's/jammy/noble/g' /etc/apt/sources.list
```
elle-j, line 86: For the [LLVM release build](https://github.com/paritytech/revive/actions/runs/19830461975/job/56814914090) that had the issue, it's using one of the `parity-large` runners and it looks like it is indeed jammy (search "os: Ubuntu 22.4.0 (jammy)" in the logs, should be in the Rust toolchain setup step). It may be preferable to do an upgrade instead?

PR #433, opened 2025-12-15
Emit `consume_all_gas` in invalid and bounds checks

### LINKED ISSUE #374: The `WordToPointer` `RuntimeFunction` implementation should consume all gas on OOB
Currently this is just a trap.

- Add a return flag to `pallet-revive` to allow reverts that consume all gas.
- Change the code generation to emit a return with said flag (instead of emitting a `llvm.trap` intrinsic).

This will for example make the `INVALID` consume all gas.

xermicus:
Closes #374

### DIFF
--- a/crates/llvm-context/src/polkavm/context/function/runtime/revive.rs
+++ b/crates/llvm-context/src/polkavm/context/function/runtime/revive.rs
@@ -51,11 +51,11 @@ impl RuntimeFunction for WordToPointer {
         )?;
 
         let block_continue = context.append_basic_block("offset_pointer_ok");
-        let block_trap = context.append_basic_block("offset_pointer_overflow");
-        context.build_conditional_branch(is_overflow, block_trap, block_continue)?;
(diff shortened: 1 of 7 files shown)


crates/llvm-context/src/polkavm/context/function/runtime/revive.rs
```diff
@@ -51,11 +51,11 @@ impl RuntimeFunction for WordToPointer {
         )?;
 
         let block_continue = context.append_basic_block("offset_pointer_ok");
-        let block_trap = context.append_basic_block("offset_pointer_overflow");
-        context.build_conditional_branch(is_overflow, block_trap, block_continue)?;
+        let block_invalid = context.append_basic_block("offset_pointer_overflow");
```
kvpanch, line 54: are there tests that check this code is generated /and executed ?

xermicus: Actually no, there seems to be no dedicated test. Nice catch thanks. #435

PR #434, opened 2025-12-15
Update code base to use LLVM 21

kvpanch:


kvpanch:
Still draft as I need to fix polkavm's dwarf handling.

.github/actions/get-llvm/action.yml
```diff
@@ -51,56 +51,73 @@ runs:
...
-                });
-                if (!asset){
-                  core.setFailed(`Artifact for '${target}' not found in release ${llvmRelease.tag_name} (${llvmRelease.html_url})`);
-                  process.exit();
-                }
-                return asset.browser_download_url;
+          // Sort all releases by publication date (newest first)
+          allReleases.sort((a, b) => {
+            return (a.published_at < b.published_at) ? 1 : ((a.published_at > b.published_at) ? -1 : 0);
+          });
```
elle-j: Just for readability.
```suggestion
          allReleases.sort((a, b) => new Date(b.published_at) - new Date(a.published_at));
```

crates/llvm-builder/src/platforms/aarch64_linux_musl.rs
```diff
@@ -221,7 +221,6 @@ fn build_host(
                 "-DCOMPILER_RT_BUILD_MEMPROF='Off'",
                 "-DCOMPILER_RT_BUILD_ORC='Off'",
                 "-DCOMPILER_RT_DEFAULT_TARGET_ARCH='aarch64'",
-                "-DCOMPILER_RT_DEFAULT_TARGET_ONLY='On'",
```
elle-j, line 224: Why does this need to be removed? Doesn't this ensure it only tries to build for `aarch64`?

crates/llvm-builder/src/platforms/x86_64_linux_musl.rs
```diff
@@ -219,7 +219,6 @@ fn build_host(
                 "-DCOMPILER_RT_BUILD_MEMPROF='Off'",
                 "-DCOMPILER_RT_BUILD_ORC='Off'",
                 "-DCOMPILER_RT_DEFAULT_TARGET_ARCH='x86_64'",
-                "-DCOMPILER_RT_DEFAULT_TARGET_ONLY='On'",
```
elle-j, line 222: Same as [previous comment](https://github.com/paritytech/revive/pull/434#discussion_r2740598911).

crates/llvm-builder/src/platforms/aarch64_linux_musl.rs
kvpanch: Unfortunately, I have no good answer to that question. That define caused compiler-rt to look for some asan  headers that in the wrong location.
I was trying to understand logic in cmake files of compiler-rt, but just gave up.
jfyi: In official docs on how to build compiler-rt, this flag is not used.

elle-j: Ah ok, strange. Is the size of the build artifact affected? Maybe the change is negligible.

xermicus:
> perhaps we should add an "Internal" section to the CHANGELOG

Good point of pointing out that it should be added to the changelog. But WDYM with "internal" section, he can just add it on top, where all new (unreleased) changes go?

crates/llvm-builder/src/platforms/aarch64_linux_musl.rs
xermicus: I can't follow, this flag tells the build system to only built the runtime for the default target (for the _host compiler_). How would it change the size of our contract builds?

>  I have no good answer to that question

Fair. They set this flag in the era compiler llvm builder and I didn't bother to investigate it first hand. It must've something to do with how cross builds work. I'll briefly look into it too.

crates/llvm-context/src/polkavm/context/mod.rs
```diff
@@ -1298,7 +1292,7 @@ impl<'ctx> Context<'ctx> {
                 call_site_value.add_attribute(
                     inkwell::attributes::AttributeLoc::Param(index as u32),
                     self.llvm
-                        .create_enum_attribute(Attribute::NoCapture as u32, 0),
+                        .create_enum_attribute(Attribute::Captures as u32, 0), // captures(none)
```
xermicus, line 1295: This caught my eye. Since `NoCapture` no longer exist. I guess the comment intents to say that it is still like no capture?

xermicus:
This change should be mentioned under `Changed`.

crates/llvm-context/src/polkavm/context/mod.rs
kvpanch: `nocapture` now is represented as `captures(none)` in textual LLVM IR

kvpanch:
> * rust-musl-cross

do you mean as a part of this PR or follow up ?

xermicus:
> > * rust-musl-cross
> 
> do you mean as a part of this PR or follow up ?

Done with #466

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ This is a development pre-release.
 
 - npm package: The `--bin` CLI flag was not producing bytecode because the `outputSelection` was not updated to explicitly requested it.
 
+### Changed
+
+- Updated `LLVM` from `18.1.8` to LLVM `21.1.8`
+- Updated `musl` to `1.2.5`
+- Updated `inkwell` to `0.8.0`
+- Updated `polka-*` to `0.31.0`
```
xermicus:
```suggestion
- Updated `polkavm-*` to `0.31.0`
```

Issue #263, opened 2025-03-20
Using the output of calldataload as a return value causes the execution to fail.


Subway2023:
Version: 0.1.0-dev.12
```solidity
contract C {
    function f(bytes memory data) public pure returns (uint256, bytes memory) {
        uint256 result1;
        assembly {
            let ptr := mload(add(data, 0x20))
            result1 := calldataload(ptr)
        }
        return (result1, data);
    }
}
```
## Run with resolc
I have set up an execution [environment ](https://github.com/Subway2023/revive)that can directly execute the bytecode generated by revive compilation.
### Compile
```
resolc --solc solc test.sol --combined-json hashes,bin,abi -o /tmp/tmpdh_0t_pr --overwrite
```
### Run
```
[...]

PR #438, opened 2026-01-01
Added 2 options to control expected result to `revive-runner`

kvpanch:
Added 2 options to control expected result to `revive-runner`:
```
      --verify-failure <VERIFY_FAILURE>
          The expected call result of the contract execution

          Possible values:
          - instantiate: The contract instantiation must fail
          - call:        The contract execution must fail

      --verify-call-gas-consumed <VERIFY_CALL_GAS_CONSUMED>
          The expected amount of consumed gas by the contract execution
```

### DIFF
--- a/crates/integration/src/tests.rs
+++ b/crates/integration/src/tests.rs
@@ -42,7 +42,6 @@ test_spec!(address, "Context", "Context.sol");
 test_spec!(value, "Value", "Value.sol");
 test_spec!(create, "CreateB", "Create.sol");
 test_spec!(call, "Caller", "Call.sol");
-test_spec!(calldata, "TestCalldata", "Calldata.sol");
 test_spec!(balance, "Balance", "Balance.sol");
 test_spec!(return_data_oob, "ReturnDataOob", "ReturnDataOob.sol");
 test_spec!(immutables, "Immutables", "Immutables.sol");


xermicus:
`success` here means the transaction didn't revert.

The "`Result`" from pallet-revive is `Ok` but the transaction was reverted. We define success as follows: The transaction was executed (requires the `Ok` variant you are seeing) _and_ it didn't revert. Everything else is considered "not successful". This is abstracted over EVM and PVM.

kvpanch:
> (The code slightly differs from #263 and that nuance is likely the culprit)

hm... indeed bringing back assert and running `make test-integration` does not produce expected crash. But if I run
```
resolc --solc solc crates/integration/contracts/Calldata.sol --combined-json hashes,bin,abi -o /tmp/tmpdh_0t_pr --overwrite
cat /tmp/tmpdh_0t_pr/combined.json
RUST_BACKTRACE=1 RUST_LOG=trace revive-runner -b 50564d0000ba0a000000000000010700c11000c0008004808407000000000e0000001c0000002a00000039000000440000005600000063616c6c5f646174615f636f707963616c6c5f646174615f6c6f616463616c6c5f646174615f73697a65636f6e73756d655f616c6c5f6761737365616c5f72657475726e7365745f696d6d757461626c655f6461746176616c75655f7472616e7366657272656405110286170463616c6c8623066465706c6f79068a061c0288b31a001f00450091009c00a800df00c801500278029702e302cb03440459048104d504e004f60425053b055e05af051706230630066006a8079511f07b10087b158475010a02013d07000002510507501002095010048e009511a07b10587b15509515608411e049111849111049018000330740641849110850100644074911384911304911284911208317200a0601821730821838821928821a20d49808d4a707d487075107123308100002838833070133090a0401330710000283778000330833090a501008a40533078000330850100adf0783783307330933000c0a0401951160fe7b1098017b1590017b1688019515a0018411e064164926780149267001492660018000330740956860014926680150100eaa067b6778390700000253170469019517e08477e07b67706471837733080a010182677080771c97772014080000000097b6da89ab873f0139070000025417233d019517e08477e07b6770647183773308040a0101826970829318829a10829b08d43a0788b801947851081801829c380700000295c823d8c809c89b027b6b48d8b20bda990bc8ab0b7b6a50d8ab097b6358c83909d4b909d87807942794975107e9009517e08477e07b677064717b6c6883c80483770a0101826c7082c91882ca1082cb08d49a0788b80194785108a1007b6a207b692882c77b6760956740013308407b6b7050101087048260580195670001827c50827b48827a4082686095871fd88708826970c898088488018477e095793fd87907c887078499e0c87b08c89a027b6a30d8a209c89803d3b30a7b6b38d8b30bdaa90bc8bc0a7b6c40d8ca0cc8c0098ea78e98da9708d49a048e37db480752071bd30907d80908db780c826840d3a808d48707db7c0b510b30501012ab02330810000228163308100002280f33081000022808330810000201838833070133093300140a04017b6220017b6328017b6a3001330740956820017b6938017b6018501016f204826730826940d49707826838826a18d4a808d48707987720d49808978820d4870752073802826760956900017b978267707b97088267207b9710956800018267287b9718826730501018a604826868826760c88707d88708826b48826970c8b909c88909d3b90ad8b90bdaa80b826a50826820c8a808c88b0bd8a80ad88b02826c58826828c88c0cc8ca0ac82a08957a24d87a07c8790cd89c09da77093807000002c8b909d8b90bc8b808d8a70788ca01857701dbca07889a01888b01da8a0bd49808db8b0752070b33081000022806ff826920826760d49707826828826a70d4a808d48707987720d49808978820d4870752076701826718826840d487071409e0ffffff000000008268307b6978d8980882693894989478510840018267688377247b676882673083772082686050101ab00483778168608269680a01826730826260c87202d87208826b38826970c8b909c88909d3b90ad8b90bdaa80b826a40826820c8a808d8a80a826c18826728c87c0cc8ca0ac88b0bd88b08c8a808d4b808826778d8720a949a948a510acc008327204926f8004926f0004926e8009568e0004926e00050101c45039567c000826830826938826a40826b1850101e0a018267d8007b67788267d0007b67708267c8007b67688267c0007b67609567a000330840501020ce018267b8008268b0008269a800826ba000d49707d48b09d47909989920d48707977720d4970752074a64ba8267607b6780008267687b6788008267707b679000956880008267787b67980064b77b6a58501022b4023308208267585010249b03837833092033070a043300260a0328083300260a03019511b07b10487b15409515508411f0491130491128491120140700000000717b484e9518207b173833075010286402491118491110490141330704641849110850102a4e02390804000256183f0b200304000240013308100002838833070133092433002c0a04019511a07b10587b15507b16489515608411e06416648c958820d8c80cc8c902d89209dacc09c8a909d8a90ac8ba0ad4980bd4a20ad4ab0b98bb20d4a909979920d4b9097b671852096d95672050102ea000826738826930826a28826820951be084bbe064b1d4a707d4980ad47a0a98aa20d49707977720d4a70752073983b783887b6b100a01826a1082a71882a81082a90882aa826b187bb7187bb8107bb9087bba9551a082105882155082164895116032003300300a03019511f87b103307501032e2f99511f87b10330701501034d5f93307080002390808000251080e330710000383770a05320032009511f07b10087b156489647533082064975010361b027c78017c797c7a027c7b03978808d4980897aa1097bb18d4ba0ad4a8087c79057c7a047c7b067c7c07979908d4a90997bb1097cc18d4cb0bd4b909979920d489027c79097c7a087c7b0a7c7c0b979908d4a90997bb1097cc18d4cb0bd4b9097c7a0d7c7b0c7c7c0e7c780f97aa08d4ba0a97cc10978818d4c808d4a808978820d498037c78117c7a107c7b127c7c13978808d4a80897bb1097cc18d4cb0bd4b8087c7a157c7b147c7c167c791797aa08d4ba0a97cc10979918d4c909d4a909979920d4890a7c78197c79187c7b1a7c7c1b978808d4980897bb1097cc18d4cb0bd4b8087c791d7c7b1c7c7c1e7c771f979908d4b90997cc10977718d4c707d49707977720d487076f776fa86f396f2a7b5a187b59107b58087b57821008821595111032009511d87b10207b15187b161082897b19088289087b19828510828618330820501038d3006f686f59821a6faa821b086fbb787b18787a10787908787898bc38787c1f98bc30787c1e98bc28787c1d98bc20787c1c98bc18787c1b98bc10787c1a98bb08787b1998ab38787b1798ab30787b1698ab28787b1598ab20787b1498ab18787b1398ab10787b1298aa08787a11989a38787a0f989a30787a0e989a28787a0d989a20787a0c989a18787a0b989a10787a0a9899087879099889387879079889307879069889287879059889207879049889187879039889107879029888087878018210208215188216109511283200838951092e8b7a10520a34330b010001aeb92cc8780883881f8488e0563800000120390a040002ae8a093d080400020183773308100002c8870732000049694884244912292149d2244992506a881aa250a5111191228244848490a46a92002284249526495214922449922415494a6a92244552220911499224499224499224492a2949922409a150282ca5119188842449922491284922494892244992244992244912922449922421499224492249004952912449283949922449922449922415898888104912229148241222229224494a241291282109a55269922449004921494a08110c25699224952449922449928424494a9224a9942425494992d254480921245425550a294992244992244992244992244992244992244992244992244992244992244992aa24a594242525095529494992244992244992244992244992244992244992244952494292042161210500 -c d45754f800000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000001a300000000000000000000000000000000000000000000000000000000000000
``` 

i.e. `-b` contains `bin-runtime`, I do see
```
assertion `left == right` failed: contract execution result mismatch: Exec { result: ContractResult { weight_consumed: Weight { ref_time: 14950859, proof_size: 2746 }, weight_required: Weight { ref_time: 14950859, proof_size: 2746 }, storage_deposit: StorageDeposit::Charge(0), max_storage_deposit: StorageDeposit::Charge(0), gas_consumed: 0, result: Ok(ExecReturnValue { flags: REVERT, data: [] }) }, wall_time: 383.167µs }

```

xermicus:
Assuming the blob you run is compiled from the `Calldata.sol` Solidity in this PR, the provided calldata is wrong. `0x89dab697` is the ABI function selector for `function bad_func_dispatch`. To me it seems the revert is correct and expected here or am I missing something?

I think it's a bit confusing that the runner automatically inserts an verification assertion there. I didn't spend much time on the runner yet. We could change that and end the execution with a more friendly error message.

> it seems essential that error code will be set since EVM does always complete successfully

Sorry I can't follow here :D What "error code" will be set where?

FWIW I don't think we need to emulate this Event behavior in `revive-runner`. I mean if the EVM always returns success, isn't the `Ok(revert)` result variant we get from the `bare_call` already close to that? We check the same `return_value.did_revert()` anyways and derive our success value from it don't we? Unless I am missing something, it is in a sense a simpler version of the `Err(<Error<T>>::ContractReverted.into());` `Error` (I see no information added by using the latter).

kvpanch:
Discussion offline with @xermicus: it's my misunderstanding of what `self.success` means, so after all the behavior of the runner is correct, but just lacks of CL controls that now are added to this PR

crates/runner/src/lib.rs
```diff
@@ -116,15 +116,15 @@ impl ExtBuilder {
     }
 }
 
-/// Expectation for a call
+/// Expectation for a call. These struct is initialized by the user and compared to the actual call result
```
xermicus:
```suggestion
/// Expectation for a call. This struct is initialized by the user and compared to the actual call result.
```

crates/runner/src/main.rs
```diff
@@ -1,8 +1,18 @@
+use std::collections::HashSet;
 use std::path::PathBuf;
 
 use clap::Parser;
 
-use revive_runner::{Code, OptionalHex, Specs, SpecsAction::*, TestAddress};
+use revive_runner::{Code, OptionalHex, Specs, SpecsAction::*, TestAddress, VerifyCallExpectation};
+
+/// The action which is expected to fail
```
xermicus:
```suggestion
/// The action which is expected to fail.
///
/// If neither is provided (default), instantiation and call are both expected to succeed.
```

xermicus (approved):

PR #451, opened 2026-01-13
Add default values to compiler arguments

### LINKED ISSUE #446: Display default value in --help
We should define the default values directly in the clapped parsed arguments.

This way they will show up when you run the help command

right now for example heap-size is an optional

```rust
    #[arg(long = "heap-size")]
    pub heap_size: Option<u32>
```

and you have to go dig the default value in the code or in the md book, it would be much nicer to discover the default value in the printed text of the --help command

elle-j:
## Description

Adds default values to the compiler arguments below when parsed by `clap`. This does not change previous user-facing behavior as these defaults were implemented using different logic, but improves the UX and the visibility of the exact defaults used.

**Example output from `resolc --help`:**
(Only affected arguments are shown here, and some of their descriptions have been omitted here.)
```
Options:

  -O, --optimization <OPTIMIZATION>
          Set the optimization parameter -O[0 | 1 | 2 | 3 | s | z]. Use `3` for best performance and `z` for minimal size
          
          [default: z]

      --metadata-hash <METADATA_HASH>
          Set the metadata hash type. Available types: `none`, `ipfs`, `keccak256`
          
          [default: keccak256]

      --heap-size <HEAP_SIZE>
          The emulated EVM linear heap memory static buffer size in bytes.
          ...
          
          [default: 131072]

      --stack-size <STACK_SIZE>
          The contracts total stack size in bytes.
          ...
          
          [default: 131072]
```

## Resolved Issues

Closes #446

### DIFF
--- a/crates/resolc/src/resolc/arguments.rs
+++ b/crates/resolc/src/resolc/arguments.rs
@@ -361,14 +361,14 @@ impl Arguments {
                     None,
                 ));
             }
-            if Self::is_explicit_argument("optimization", &argument_matches) {
+            if Self::is_argument("optimization", &argument_matches) {
                 messages.push(SolcStandardJsonOutputError::new_error(
                     "LLVM optimizations must be specified in standard JSON input settings.",
                     None,
                     None,
                 ));
             }
-            if Self::is_explicit_argument("metadata_hash", &argument_matches) {
+            if Self::is_argument("metadata_hash", &argument_matches) {
                 messages.push(SolcStandardJsonOutputError::new_error(
                     "Metadata hash mode must be specified in standard JSON input settings.",
                     None,
@@ -376,14 +376,14 @@ impl Arguments {
                 ));
             }
 
-            if Self::is_explicit_argument("heap_size", &argument_matches) {
+            if Self::is_argument("heap_size", &argument_matches) {
                 messages.push(SolcStandardJsonOutputError::new_error(
                     "Heap size must be specified in standard JSON input polkavm memory settings.",
                     None,
                     None,
                 ));
             }
-            if Self::is_explicit_argument("stack_size", &argument_matches) {
+            if Self::is_argument("stack_size", &argument_matches) {
                 messages.push(SolcStandardJsonOutputError::new_error(
                     "Stack size must be specified in standard JSON input polkavm memory settings.",
                     None,
@@ -416,7 +416,7 @@ impl Arguments {
     /// argument, even if it is the same value as the default one.
     ///
     /// Panics if the `id` is not a valid argument id.
-    fn is_explicit_argument(argument_id: &str, argument_matches: &ArgMatches) -> bool {
+    fn is_argument(argument_id: &str, argument_matches: &ArgMatches) -> bool {
         argument_matches
             .value_source(argument_id)
             .is_some_and(|source| source == ValueSource::CommandLine)


crates/resolc/src/resolc/arguments.rs
```diff
@@ -403,6 +410,19 @@ impl Arguments {
...
+    ///
+    /// This can be used for determining if a user has explicitly set a value for
+    /// an argument, even if it is the same value as the default one.
+    ///
+    /// Panics if the `id` is not a valid argument id.
+    fn came_from_default_value(argument_id: &str, argument_matches: &ArgMatches) -> bool {
+        argument_matches
+            .value_source(argument_id)
+            .is_some_and(|source| source == ValueSource::DefaultValue)
+    }
```
elle-j: For reference and examples, see [value_source](https://docs.rs/clap/latest/clap/struct.ArgMatches.html#method.value_source) and [default_value](https://docs.rs/clap/latest/clap/struct.Arg.html#method.default_value).

crates/resolc/src/cli_utils.rs
```diff
@@ -158,6 +160,22 @@ pub fn assert_command_failure(result: &CommandResult, error_message_prefix: &str
...
+            .any(|error| error.is_error() && error.message.contains(error_message)),
+        "the standard JSON output should contain the error message `{error_message}`"
+    );
+}
+
+/// Converts valid JSON text to `SolcStandardJsonOutput`.
+pub fn to_solc_standard_json_output(json_text: &str) -> SolcStandardJsonOutput {
+    serde_json::from_str(json_text).unwrap()
+}
+
```
xermicus: Since these are only used in tests please move them to the tests

crates/resolc/src/resolc/arguments.rs
```diff
@@ -403,6 +409,19 @@ impl Arguments {
         messages
     }
 
+    /// Checks whether the value from the argument with the given `argument_id`
+    /// came from the preconfigured default value.
+    ///
+    /// This can be used for determining if a user has explicitly set a value for
```
xermicus:
```suggestion
    /// This can be used for determining if a user has explicitly provided
```

Nit

crates/resolc/src/resolc/main.rs
```diff
@@ -170,20 +167,16 @@ fn main_inner(
...
-    };
+    let mut optimizer_settings = OptimizerSettings::try_from_cli(arguments.optimization)?;
     optimizer_settings.is_verify_each_enabled = arguments.llvm_verify_each;
     optimizer_settings.is_debug_logging_enabled = arguments.llvm_debug_logging;
 
-    let metadata_hash = match arguments.metadata_hash {
-        Some(ref hash_type) => MetadataHash::from_str(hash_type.as_str())?,
-        None => MetadataHash::Keccak256,
-    };
+    let metadata_hash = arguments.metadata_hash;
```
xermicus:
```suggestion
    let metadata_hash = arguments.metadata_hash;
```

Nit: The binding is no longer necessary.

crates/resolc/src/resolc/arguments.rs
```diff
@@ -403,6 +409,19 @@ impl Arguments {
...
+    /// an argument, even if it is the same value as the default one.
+    ///
+    /// Panics if the `id` is not a valid argument id.
+    fn came_from_default_value(argument_id: &str, argument_matches: &ArgMatches) -> bool {
```
xermicus:
```suggestion
    fn is_argument(argument_id: &str, argument_matches: &ArgMatches) -> bool {
```

Nit

xermicus (approved):
LGTM modulo some nits

crates/resolc/src/resolc/arguments.rs
elle-j: To me `is_argument` doesn't differentiate enough between if it was implicit or explicit. We could update the function to either `is_explicit_argument` or `is_implicit_argument` though.

xermicus: Well there is no such thing as "implicit arguments". They are either provided by the user or not. `is_argument` makes sense because if `--standard-json` is an argument, we check `is_argument` for the ones we don't want to allow. Anyways, it's not terribly important.

PR #453, opened 2026-01-14
ci: Add build tool to workflow

elle-j:
Fixes [this](https://github.com/paritytech/revive/actions/runs/21000852574/job/60370115169) and [this](https://github.com/paritytech/revive/actions/runs/21008888013/job/60398447550) failed runs.


.github/workflows/benchmark.yml
```diff
@@ -28,21 +28,28 @@ jobs:
     permissions:
       pull-requests: write
     steps:
+      - name: Checkout Commit Before Changes
```
xermicus, line 31: Out of curiosity: Why does this need to be on top?

elle-j: Mainly for clarity to see at the top what's being checked out. Minor thing tho

PR #454, opened 2026-01-15
Update rendering of `resolc` binaries Markdown tables

elle-j:
Fixes the rendering of the Markdown tables for `resolc-bin` which do not render inside the `details` tags for GH pages. Removes `details` and instead adds a table of contents at the top for fast navigation to respective platform.


xermicus (approved):
Thanks! Unfortunate that my MD preview rendered the tables in spoilers just fine but apparently not all MD engines are that good.

PR #455, opened 2026-01-15
remove superfluous byte swap in `linkersymbol` builtin translation

xermicus:
- Remove the superfluous byte swap in the `linkersymbol` builtin translation.
- Support testing the `linkersymbol` builtin translation in the runner specs.

### DIFF
--- a/crates/llvm-context/src/polkavm/context/mod.rs
+++ b/crates/llvm-context/src/polkavm/context/mod.rs
@@ -748,7 +748,7 @@ impl<'ctx> Context<'ctx> {
         &self,
         pointer: Pointer<'ctx>,
     ) -> anyhow::Result<inkwell::values::BasicValueEnum<'ctx>> {
-        let address = self.build_byte_swap(self.build_load(pointer, "address_pointer")?)?;
+        let address = self.build_byte_swap(self.build_load(pointer, "address_value")?)?;
         Ok(self
             .builder()
             .build_int_z_extend(address.into_int_value(), self.word_type(), "address_zext")?


crates/llvm-context/src/polkavm/evm/call.rs
```diff
@@ -163,7 +163,16 @@ pub fn linker_symbol<'ctx>(
         context.integer_type(revive_common::BIT_LENGTH_ETH_ADDRESS),
         Default::default(),
     );
-    context.build_load_address(context.get_global(path)?.into())
+
+    let symbol = context.get_global(path)?;
+    let value = context.build_load(symbol.into(), path)?;
```
elle-j, line 168: Do we want to keep the "_pointer" here as in the [previous function](https://github.com/paritytech/revive/blob/2d85ca9755adcebb5815c0323cfa7985b2aca227/crates/llvm-context/src/polkavm/context/mod.rs#L751)?
```suggestion
    let value = context.build_load(symbol.into(), &format!("{path}_pointer"))?;
```

elle-j, line 168: Nit: `address` instead of `value` may make it clearer.

xermicus: No. That was the reason for the bug first hand. It's not an address but just bytes.

elle-j: Ah okay 🙂

xermicus: Why would we want to label a value as a pointer?

xermicus: FWIW looks like the name is nonsensical in the place you pointed out. I fixed it in the address loading helper as a drive-by. Thanks for pointing this out!

elle-j: That suggestion was made when I still saw it as an address 👍

xermicus: Sorry can't follow, the result of this load is a value anyways, regardless whether conceptually we consider it an address or just bytes. The suffix would be `_value`. However, while I think having having `_pointer` suffixes can ease following indirections, for values I don't think it matters all that much. I'd like to keep it here as is (I think for clarity, if anything, the suffix should be something like `_linker_symbol`, but let's not bikeshed this).

PR #459, opened 2026-01-20
Add integration test for deleting struct

elle-j:
Differentially tests the behavior of deleting a struct with a member that do not occupy the full 32-byte slot. From [this issue](https://github.com/paritytech/contract-issues/issues/118) it was unclear whether the `delete` was expected to reset only the bytes occupied or the entire slot. The current behavior of resetting the entire slot is expected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9155,13 +9155,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "revive-replay"
-version = "0.1.0"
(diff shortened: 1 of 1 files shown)


Cargo.lock
elle-j, line 1: This was updated automatically.

xermicus: Yes looks like a left-over artifact was accidentially added.

PR #461, opened 2026-01-26
Improve compilation time in standard JSON mode and honor requested output

### LINKED ISSUE #326: Extremely long compilation with `resolc` when lots of files are involved.
### Description 
Compile times with `resolc` are way longer than they are with `solc` on an example project
[Project link - Centrifuge/protocol-v3](https://github.com/centrifuge/protocol-v3/tree/main)

#### standard-json with test and script files
[standard-json.json](https://github.com/user-attachments/files/20373360/standard-json.json)
 - `solc v0.8.28`: 57s
 - `resolc v0.1.0-dev.16 and solc v0.8.28`: 2647s and failed compilation 
     - this is probably related to this issue as the code that causes the compilation error uses public functions [see](https://github.com/paritytech/revive/issues/91)

#### standard-json without test and script files

[standard-json.json](https://github.com/user-attachments/files/20373473/standard-json.json)

 - `solc v0.8.28`: 6s
 - `resolc v0.1.0-dev.16 and solc v0.8.28`: 36s

elle-j:
# Description

Compilation times for resolc were significantly slower than solc when:
* Multiple files were included in the JSON input file via `--standard-json`
* And none of the files requested code generation
* (If only a subset of files requested code generation, both resolc and solc were very slow)

Reason:
* Resolc unconditionally generated bytecode for all contracts, even if not requested.
* When compiling to PolkaVM bytecode, we need to pass additional flags (e.g. `irOptimized`) to solc. If these flags were not explicitly requested, resolc also added these to the requested output since PolkaVM bytecode was always generated.

## Breaking Changes

The following changes are breaking, but fix the semantics to match its intended behavior:

* Compiling with `--standard-json` now compiles to PolkaVM bytecode ***only*** for the contracts explicitly requested in the `outputSelection`.
* Compiling with `--standard-json` now outputs ***only*** the explicitly requested output (e.g. `evm.assembly`, `evm.bytecode`, `ast`, etc.).

```jsonc
// Example:
{
  "settings": {
    // ...
    "outputSelection": {
      "src/common/GasService.sol": {
        // Contracts in this file will generate bytecode.
        // Only these fields of the JSON output selection will be in the `contracts` output.
        "*": ["abi", "evm.methodIdentifiers", "metadata", "evm.bytecode"],
        // Only this field of the JSON output selection will be in the `sources` output.
        "": ["ast"]
      },
      "src/common/Gateway.sol": {
        // No contracts in this file will generate bytecode.
        "*": ["abi", "evm.methodIdentifiers", "metadata"],
        // No `ast` will be in the `sources` output (only the automatically added `id`,
        // similar to solc as this is not a configurable output selection).
        "": []
      },
    }
  }
}
```

## Performance Improvements

When e.g. the standard JSON input file includes multiple files and requests output for every contract as per this setting:

```jsonc
{
  "settings": {
    // ...
    "optimizer": { "enabled": true, "runs": 1 },
    "viaIR": true,
    "outputSelection": {
      "src/common/GasService.sol": {
        "*": ["abi", "evm.methodIdentifiers", "metadata", "evm.bytecode"],
      },
      // Rest of files..
    }
  }
}
```

The differences in compilation times depending on whether `evm.bytecode` is selected in every file, one file, or no files can be seen below:

| Solc Optimizer (runs=1) | ViaIR | Request `abi`, `evm.methodIdentifiers`, `metadata` | Request `evm.bytecode` | resolc before | resolc after | solc v0.8.33      |
|:-----------------------:|:-----:|:---------:|:---------:|:-------------:|:------------:|:----------:|
| ✅                      | ✅    | All files | All files | `~14.5s`     | `~14.5s`    | `~13.5s`  |
| ✅                      | ✅    | All files | One file  | `~14.5s`     | `~450ms`    | `~13.5s`  |
| ✅                      | ✅    | All files | No files  | `~14.5s`     | `~390ms`    | `~350ms`  |

**Observations:**

* `resolc before` is roughly the same no matter the request
  * This is due to unconditionally generating bytecode
* When not requesting bytecode for any contracts:
  * `resolc after` now compiles in ~390ms (vs ~14.5s), similar to `solc`
  * Neither `solc` nor `resolc after` generates bytecode when none of the contracts request it
* When requesting bytecode for all contracts in _only one_ file:
  * `resolc after` now compiles in ~450ms (vs ~14.5s)
  * `solc` still compiles in the same time as when requesting bytecode for all contracts in all files
    * This is something they have tried to improve (see [Solidity#15373](https://github.com/argotorg/solidity/issues/15373) and [Solidity#15615](https://github.com/argotorg/solidity/issues/15615)), and the intended behavior seems to now be that bytecode should only be generated when explicitly requested for each contract.
    * Even though PRs have merged to try to fix it, their issue seems to persist as of v0.8.33.

# Open Questions

* Foundry:
  * [x] Pruning of the AST:
    * Just saw [this past commit](https://github.com/paritytech/revive/pull/385) to not prune the AST for Foundry compatibility.
    * Can Foundry not explicitly request `"": "ast"`?
    * **Answer:**
      * Pruning the AST when not requested should be fine, as long as Foundry explicitly requests it when needed.
      * Will be tested by Foundry maintainer.

# Review Notes

Changes in `docs/` (~25 files) can be skipped as they are automatically updated after adding content to `book/` and rebuilding it.

# Resolved Issues

Closes #326

### DIFF
--- a/crates/solc-json-interface/src/standard_json/input/settings/selection/mod.rs
+++ b/crates/solc-json-interface/src/standard_json/input/settings/selection/mod.rs
@@ -52,11 +52,6 @@ impl PerFileSelection {
         None
     }
 
-    /// Checks whether any of the `flags` is selected in any of the files.
-    pub fn contains_any(&self, flags: &[Flag]) -> bool {
-        self.files.values().any(|file| file.contains_any(flags))
-    }
-
     /// Checks whether this is the empty per file selection.
     pub fn is_empty(&self) -> bool {
         self.files.is_empty()


crates/solc-json-interface/src/standard_json/input/settings/selection/file/flag.rs
```diff
@@ -49,12 +49,53 @@ pub enum Flag {
...
+    }
+
+    /// Whether this selection flag is specific for the solc backend only.
     ///
     /// Specifically, EVM bytecode and related flags should never be requested.
     /// It will be replaced by PVM code anyways.
-    pub fn is_required_for_codegen(&self) -> bool {
-        !matches!(
+    pub fn is_solc_backend(&self) -> bool {
+        matches!(
```
elle-j, line 98: As codegen is no longer unconditionally done, this rename better reflects whether or not something should be passed to solc. This method was called by `retain()` methods deciding what to retain in the output selection before being passed to solc.

crates/solc-json-interface/src/standard_json/input/settings/selection/file/mod.rs
```diff
@@ -62,44 +68,66 @@ impl File {
...
     }
 
-    /// Returns flags that are going to be automatically added by the compiler,
-    /// but were not explicitly requested by the user.
+    /// Returns flags that were not explicitly requested by the user.
     ///
-    /// Afterwards, the flags are used to prune JSON output before returning it.
+    /// These flags are used to prune JSON output before returning it,
+    /// removing any output that was automatically added but not requested.
     pub fn selection_to_prune(&self) -> Self {
```
elle-j, line 75: Previously, this function only added a subset of flags to the selection to prune (the ones automatically added that were needed for codegen, e.g. `irOptimized` etc.). Since bytecode was generated unconditionally, other output fields (e.g. `evm.bytecode`) were still in the output even if it was not requested. This updates the function to add all flags that were not requested to the selection to prune.

.github/workflows/benchmark.yml
```diff
@@ -95,7 +95,7 @@ jobs:
           cargo bench --package resolc --bench compile -- --save-baseline resolc_changes
           cargo bench --package revive-benchmarks --bench parse -- --save-baseline yul_parse_changes
           cargo bench --package revive-benchmarks --bench lower -- --save-baseline yul_lower_changes
-        timeout-minutes: 20
+        timeout-minutes: 30
```
kvpanch: why is this change necessary ?

crates/resolc/benches/compile.rs
```diff
@@ -79,7 +81,7 @@ fn bench_dependency(c: &mut Criterion) {
     let mut group = group(c, "Dependency");
     group
         .sample_size(50)
-        .measurement_time(Duration::from_secs(9));
+        .measurement_time(Duration::from_secs(10));
```
kvpanch, line 84: why is it needed ?

.github/workflows/benchmark.yml
elle-j: Due to adding a benchmark where solc will run longer.

crates/resolc/benches/compile.rs
elle-j: This actual benchmark itself should be unaffected by the changes, but we haven't rerun the benchmarks locally and updated the file for a while. For this run on my machine it now took between 9-10sec to complete the samples, since it's over 9sec, a warning is emitted (suggesting to increase it) when running the benchmarks, so I increased it 👍

crates/solc-json-interface/src/standard_json/output/source.rs
```diff
@@ -20,6 +20,7 @@ pub struct Source {
     /// The source code ID.
     pub id: usize,
     /// The source code AST.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
xermicus:
```suggestion
    #[serde(default, skip_serializing_if = "serde_json::Value::is_null")]
```
Can probably remove the `Option` and do this instead.

crates/resolc/BENCHMARKS_M4PRO.md
```diff
@@ -6,47 +6,62 @@
...
+|        | `resolc`               | `solc`                         |
+|:-------|:-----------------------|:------------------------------ |
+|        | `1.51 s` (✅ **1.00x**) | `1.11 s` (✅ **1.37x faster**)  |
+
+
+### Std JSON Codegen One of Some Files
+
+|        | `resolc`                  | `solc`                           |
+|:-------|:--------------------------|:-------------------------------- |
+|        | `454.24 ms` (✅ **1.00x**) | `13.82 s` (❌ *30.42x slower*)    |
```
xermicus: This seems sus: I'd not expect to `solc` to be so much slower here? IIRC the usual culprit of the time spent in `solc` is in the yul optimizer. But we require optimized the optimized Yul IR so it should trickle down.

crates/resolc/BENCHMARKS_M4PRO.md
```diff
@@ -6,47 +6,62 @@
...
 
-### Multiple Contracts (`--standard-json`)
+### Std JSON No Codegen Many Files
```
xermicus: It's a bit confusing to have codegen with one file and _no_ codegen with _many_ files. Can we du codegen and no codegen with one and many files?

crates/solc-json-interface/src/standard_json/input/settings/selection/file/flag.rs
```diff
@@ -49,12 +49,53 @@ pub enum Flag {
...
+            Self::EVMLA,
+            Self::EVMBC,
+            Self::EVMDBC,
+            Self::Assembly,
+            Self::Ir,
+        ]
+    }
+
+    /// Returns the EVM child flags.
+    /// Excludes the EVM legacy assembly as it does not have a PVM equivalent.
```
xermicus, line 73: This is incomplete? Looks like either the doc comment is missleading or what the this returns is wrong.

crates/solc-json-interface/src/standard_json/input/settings/selection/file/flag.rs
```diff
@@ -49,12 +49,53 @@ pub enum Flag {
...
+        ]
+    }
+
+    /// Whether this selection flag is specific for the solc backend only.
     ///
     /// Specifically, EVM bytecode and related flags should never be requested.
     /// It will be replaced by PVM code anyways.
-    pub fn is_required_for_codegen(&self) -> bool {
-        !matches!(
+    pub fn is_solc_backend(&self) -> bool {
```
xermicus:
```suggestion
    pub fn is_evm_codegen(&self) -> bool {
```

crates/solc-json-interface/src/standard_json/input/settings/selection/file/flag.rs
xermicus: Thanks for the explanation! Makes sense to me. But I'd prefer to name it `is_evm_codegen` (we don't have a solc backend and I tripped over this).

crates/resolc/BENCHMARKS_M4PRO.md
elle-j: The reason for this is in the PR description > Performance Improvements > Observations:

* When requesting bytecode for all contracts in only one file
  * `solc` still compiles in the same time as when requesting bytecode for all contracts in all files
    * This is something they have tried to improve (see [Solidity#15373](https://github.com/argotorg/solidity/issues/15373) and [Solidity#15615](https://github.com/argotorg/solidity/issues/15615)), and the intended behavior seems to now be that bytecode should only be generated when explicitly requested for each contract.
    * Even though PRs have merged to try to fix it, their issue seems to persist

crates/solc-json-interface/src/standard_json/input/settings/selection/file/flag.rs
elle-j: Hmm, well the intention here is to return `` `evm` `` child flags relevant for us (so not the `evm.legacyAssembly`). It is used for determining whether a PVM equivalent output is requested. Perhaps just updating the doc from all caps `EVM` to lower case `` `evm` `` to signify the actual `evm` field? Feel free to suggest a change based on what you find misleading.

crates/solc-json-interface/src/standard_json/input/settings/selection/file/flag.rs
elle-j: Replying to [this](https://github.com/paritytech/revive/pull/461#discussion_r2733352444):

I agree, good suggestion 🙂 I primarily meant "is intended to be used for solc's backend"

crates/solc-json-interface/src/standard_json/input/settings/selection/file/flag.rs
xermicus: The doc comment says it returns the `evm` object flags (except evmla). But when I look what at what it returns I see it returns only a subset (misses more than just evmla).

crates/resolc/BENCHMARKS_M4PRO.md
xermicus: Sorry I overlooked that! Can we add this as a comment in the benchmarking code?

crates/solc-json-interface/src/standard_json/input/settings/selection/file/flag.rs
elle-j: Out of our supported evm child flags, it returns:
```
Self::MethodIdentifiers // "evm.methodIdentifiers"
Self::EVMBC,            // "evm.bytecode"
Self::EVMDBC,           // "evm.deployedBytecode"
Self::Assembly,         // "evm.assembly"
```

Which other ones are you referring to?

crates/solc-json-interface/src/standard_json/output/source.rs
elle-j: Ah yeah, sweet. [change](https://github.com/paritytech/revive/pull/461/changes/8adfdb7a654dbf174876180cb3c54298bb6e0e8f) <--

crates/resolc/BENCHMARKS_M4PRO.md
elle-j: For sure 👍

crates/resolc/BENCHMARKS_M4PRO.md
elle-j: Yeah we could make this more cohesive.

Currently:
* The codegen for one file fixture still has multiple files, but only requests codegen for one of them.
* The no codegen for many files fixture has _lots_ of files (more than the above), and does not request codegen for any of them.

Suggestion:
* We want to test and be able to compare comp time between:
  * All files request codegen (to get max time)
  * One file (still out of many) requests codegen
    * To test that codegen is not done for all, even if it is not in the output
  * No files (still out of many) requests codegen
    * To test that no codegen is done at all, even if it is not in the output

Given that the fixture with _lots_ of files would likely require hours for benchmarks to run with solc and resolc if requesting codegen for all (and that `make test` will run each benchmark but only one single time each), I think we instead could use the above suggestion using the fixture that has multiple files but not an extreme amount of files. This would still give us the above insights.

crates/solc-json-interface/src/standard_json/input/settings/selection/file/flag.rs
elle-j: I agree, good suggestion 🙂 I primarily meant "is intended to be used for solc's backend"

crates/resolc/BENCHMARKS_M4PRO.md
xermicus: Ok let's not bikeshed this and merge the PR soon. We can revisit that topic later.

crates/solc-json-interface/src/standard_json/input/settings/selection/file/flag.rs
xermicus: There exist a bunch more `evm.*`, [see the docs for reference](https://docs.soliditylang.org/en/latest/using-the-compiler.html)

elle-j: Yes for the complete support. But this is for the resolc supported flags. I kept the same support as before this PR in terms of the flags. It can be a follow-up to support the `evm.bytecode` subfields. (I [added to the docs](https://github.com/paritytech/revive/pull/461/changes/0b53099cf296f89932269333ad69f3276d4bb7a1) btw to make that clearer.)

crates/resolc/BENCHMARKS_M4PRO.md
elle-j: I made those updates after writing the suggestion, just had to wait for benches to run. But [these](https://github.com/paritytech/revive/pull/461/changes/1668c67dedea82c82bf22402277404bae0ec8049) were the updates (replaced the huge file with the ones using less files), and made the benches more comparable.

PR #466, opened 2026-01-30
llvm 21 musl cross compilation build fixes

xermicus:
The cumulative LLVM21 musl build fixes:
- Update rust-musl-cross and Rust versions
- Bring back `DCOMPILER_RT_DEFAULT_TARGET_ONLY` to prevent multilib related issues in the cross build
- Use lld in the musl target stage
- ~~Do not build the LTO tool shared object (this caused linker errors for some reason)~~ scrapped - that seemed sus to me and I tested in a clean full-stage build. Maybe was an issue with the build cache.

### DIFF
--- a/crates/llvm-builder/src/platforms/aarch64_linux_musl.rs
+++ b/crates/llvm-builder/src/platforms/aarch64_linux_musl.rs
@@ -295,7 +295,6 @@ fn build_target(
                 "-G",
                 "Ninja",
                 "-DBUILD_SHARED_LIBS='Off'",
-                "-DLLVM_TOOL_LTO_BUILD='Off'",
                 "-DLLVM_USE_LINKER='lld'",
                 "-DLINKER_SUPPORTS_COLOR_DIAGNOSTICS=0",
                 format!(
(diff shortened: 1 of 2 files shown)


crates/llvm-builder/src/platforms/aarch64_linux_musl.rs
```diff
@@ -221,6 +221,8 @@ fn build_host(
                 "-DCOMPILER_RT_BUILD_MEMPROF='Off'",
                 "-DCOMPILER_RT_BUILD_ORC='Off'",
                 "-DCOMPILER_RT_DEFAULT_TARGET_ARCH='aarch64'",
+                "-DCOMPILER_RT_DEFAULT_TARGET_ONLY='On'",
```
kvpanch, line 224: Do you have explanation why this works now ? Due to target triple ?
I thought I tried it, but maybe got drown by uncertainty of the issue.

xermicus: Yeah, setting the C compiler target fixed it.

PR #468, opened 2026-02-02
Implement verification of reproducible builds and run differential E2E tests

### LINKED ISSUE #175: Comprehensive differential E2E tests
Differential E2E tests help us ensure ETH compatibility as well as reproducible builds across supported platforms.

A [comprehensive testing framework](https://github.com/paritytech/revive-differential-tests) is currently being developed. We need a CI workflow which uses the harness and executes implemented cases (after all other workflows succeeded) as follows:

The following `resolc` binaries should be tested per platform for reproducible builds:
- Linux MUSL
- MacOS universal
- Windows
- Wasm32 emscripten 

This means the contracts will be compiled and their code hashes collected (the tests itself are only executed on linux).

The following settings must be used to test with linux musl:
- `-O3`
- `-Oz`
- `-O0`

This ensures that all opt levels work fine.

The workflows should be triggered at late stage in the pipeline, after the musl build succeeded and use the musl build to run the tests.

elle-j:
## Description

This adds verification for reproducible builds (via [revive-differential-tests](https://github.com/paritytech/revive-differential-tests) testing framework) by:

1. Compiling [contracts](https://github.com/paritytech/resolc-compiler-tests/tree/main/fixtures/solidity) and generating hashes from each bytecode
    * With the resolc builds:
      * linux musl
      * macos universal
      * windows
      * (future todo: wasm32 emscripten)
    * With optimization levels `0`, `3`, `z` for each contract
2. Exporting the hashes
3. Comparing the hashes
    * For each optimization level:
      * Checks if each hash for a specific contract is identical across the platforms

This PR also runs the differential E2E tests (via `retester`) with the resolc linux musl build.

All of the above only runs if our `test` workflow succeeds.

## Related

This PR initially implemented the above in this repo but has now been implemented into the separate testing framework instead. The most relevant PRs implemented alongside this PR as prerequisites are:

* `revive-differential-tests`:
  * [Generate bytecode hashes included in the report
](https://github.com/paritytech/revive-differential-tests/pull/238)
  * [Fix handling of resolc's optimizer `mode` and allow setting solc optimizer enabled](https://github.com/paritytech/revive-differential-tests/pull/243)
  * [Fix solc and resolc optimization settings](https://github.com/paritytech/revive-differential-tests/pull/253)
  * [Implement a `compile` context/subcommand for standalone compilation mode](https://github.com/paritytech/revive-differential-tests/pull/239)
  * [Implement support for `--mode` CLI argument for `compile`](https://github.com/paritytech/revive-differential-tests/pull/246)
  * [Implement `export-hashes` and `compare-hashes` subcommands for report processor
](https://github.com/paritytech/revive-differential-tests/pull/250)
  * [Implement GH actions for compiling, exporting and comparing hashes](https://github.com/paritytech/revive-differential-tests/pull/251)
  * [Gate zombienet-sdk behind `cfg(unix)` to fix Windows builds](https://github.com/paritytech/revive-differential-tests/pull/254)
  * [Link library-using contracts with dummy addresses for `--compile` command](https://github.com/paritytech/revive-differential-tests/pull/267)
* `revive`:
  * [Build LLVM for Windows using MSVC-compatible Clang](https://github.com/paritytech/revive/pull/497)
  * [Apply LLVM patches for deterministic operand order and correct peephole optimization for `minsize`
](https://github.com/paritytech/revive/pull/500)
* Investigation of comparison mismatches in third-party dependencies:
  * `llvm-project`:
    * [`SelectionDAG` type legalization produces different operand ordering for `i256` multiply across platforms](https://github.com/llvm/llvm-project/issues/190148)
  * `solidity`:
    * [Windows absolute paths in standard JSON break import resolution and library linking](https://github.com/argotorg/solidity/issues/16579)

### Failed compilations

Since [resolc-compiler-tests](https://github.com/paritytech/resolc-compiler-tests) currently includes contracts that are expected to fail resolc compilation (e.g. due to unsupported instructions, contracts that are expected to work with solc for evm/revm), the compilation workflow job is not expected to fail if compilation fails for a contract.

Instead, if a contract compiles with one build but not with another, a hash will be missing and will thereby generate a mismatch when comparing all hashes, failing at the verification step.

## Resolved Issues

Closes [#175](https://github.com/paritytech/revive/issues/175)


.github/workflows/test-builds.yml
```diff
@@ -0,0 +1,177 @@
...
+          hash-artifact-names: "hashes-linux, hashes-macos, hashes-windows"
+          modes: ${{ env.MODES }}
+          upload-artifact-name: hash-comparison-result
+
+  run-differential-tests:
+    needs: build
+    runs-on: ubuntu-24.04
+    timeout-minutes: 60
+    env:
+      POLKADOT_SDK_REF: "30cda2aad8612a10ff729d494acd9d5353294d63"
```
elle-j: Using the same ref as used in the test workflow in revive-differential-tests

.github/assets/revive-dev-node-polkavm-resolc.json
elle-j, line 1: This was the expectations file (expected test execution failures) generated by retester in [this run](https://github.com/paritytech/revive/actions/runs/24132123802/job/70412879350).

Reason for the test execution failure of [`fixtures/solidity/simple/algorithm/cryptography/caesar_cypher.sol::2::Y M0 S-`](https://github.com/paritytech/resolc-compiler-tests/blob/abc15edbeb84cc69e572b413e8527b101a761c6d/fixtures/solidity/simple/algorithm/cryptography/caesar_cypher.sol):

> *"Failed to execute all of the steps on the driver: Failure on step 0: Function call step Failed: Failed to handle function call assertions: Transaction status assertion failed - Expected true but got false. Revert reason: Some(\"ContractTrapped\")"*

.github/workflows/test-builds.yml
```diff
@@ -0,0 +1,192 @@
...
+          upload-artifact-name: hash-comparison-result
+
+  run-differential-tests:
+    needs: build
+    runs-on: parity-large
+    timeout-minutes: 75
+    env:
+      # TODO: Which Polkadot SDK commit do we want to use?
+      # The below one is as of March. 25, 2026.
+      POLKADOT_SDK_REF: "9d8a2fefd2414591c4a243460f3ffc5195a71deb"
```
elle-j: This is a commit from Polkadot SDK `main`. Is there a particular tag we should use instead?

kvpanch: Should this be aligned to what we have in `Cargo.toml` ? I.e. use the one that corresponds to `2507.4.0` ?

also, @xermicus, don't we need to also update this dependency ?

xermicus: Yes we test against the exact same time everywhere

xermicus (commented):
Running differential tests needs build caches for revive dev node and rpc binaries since (--profile production builds take long time to build)

.github/workflows/test-builds.yml
elle-j: Ok I'll update to use the one that's in Cargo.toml

.github/workflows/differential-tests.yml
```diff
@@ -0,0 +1,192 @@
...
+          components: "rust-src"
+          rustflags: ""
+
+      - name: Run Differential E2E Tests
+        uses: ./revive/revive-differential-tests/.github/actions/run-differential-tests
+        with:
+          revive-differential-tests-path: revive/revive-differential-tests
+          polkadot-sdk-path: ./polkadot-sdk
+          cargo-command: "forklift cargo"
+          platform: revive-dev-node-polkavm-resolc
```
elle-j, line 163: > Running differential tests needs build caches for revive dev node and rpc binaries since (--profile production builds take long time to build)

@xermicus, answering your comment ☝️ here:

I'm using [forklift](https://github.com/paritytech/forklift) for caching, which is how the Polkadot SDK also caches. [This](https://forum.parity.io/t/cargo-caching-with-forklift/2181) is an older forum post about it as well. 

You can see the cache being used e.g. in [this run](https://github.com/paritytech/revive/actions/runs/24233869428/job/70753986723).

.github/assets/revive-dev-node-polkavm-resolc.json
elle-j: Update:

This file has now been updated with the failures from [this run](https://github.com/paritytech/revive/actions/runs/24384778406/job/71218189661?pr=468) (using polkadot-sdk sha corresponding to v2603.0.0).

.github/workflows/test-builds.yml
```diff
@@ -0,0 +1,194 @@
+name: Verify Reproducible Builds
+
+on:
+  workflow_call:
+
+env:
+  # The compiler modes to use for compilation by retester.
+  MODES: "Y M0 S+, Y M3 S+, Y Mz S+"
+  SOLC_VERSION: "0.8.33"
```
xermicus:
```suggestion
  SOLC_VERSION: "0.8.34"
```

We usually use the latest supported version on CI

.github/workflows/reusable-build.yml
```diff
@@ -185,8 +249,11 @@ jobs:
           mkdir -p solc
           curl -sSLo solc/soljson.js https://github.com/argotorg/solidity/releases/download/v0.8.34/soljson.js
           node -e "
-          const soljson = require('solc/soljson');
-          const createRevive = require('./target/wasm32-unknown-emscripten/release/resolc.js');
+          // Make sure to require './solc/soljson' and not 'solc/soljson' in order to use
+          // the explicitly downloaded solc, rather than resolving an npm package potentially
```
xermicus, line 303: Did this actually happen?

.github/workflows/reusable-build.yml
```diff
@@ -142,39 +175,70 @@ jobs:
...
       - name: Build
+        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
         run: |
           export LLVM_SYS_211_PREFIX=$PWD/llvm-x86_64-unknown-linux-gnu
           export REVIVE_LLVM_TARGET_PREFIX=$PWD/llvm-wasm32-unknown-emscripten
           source emsdk/emsdk_env.sh
           make install-wasm
           chmod -x ./target/wasm32-unknown-emscripten/release/resolc.wasm
 
+      - name: Compress Artifact
```
xermicus, line 284: Why did this move down below?

.github/workflows/differential-tests.yml
xermicus, line 1: Nit: Can we name this file more accordingly to what it does, e.g. `differential-tests.yml`, we already have a lot of CI files and `test-builds` isn't descriptive enough

.github/workflows/test-builds.yml
```diff
@@ -0,0 +1,194 @@
...
+
+jobs:
+  build:
+    uses: ./.github/workflows/reusable-build.yml
+    with:
+      is_release: false
+      retention_days: 1
+
+  create-macos-universal:
```
xermicus: Grepping for "lip resolc" in .github already reveals 2 search results. Could we factor this part here (job creating the universal mac binary) out and then use it everywhere instead?

.github/workflows/test-builds.yml
```diff
@@ -0,0 +1,194 @@
...
+  compile-and-export-hashes-native:
+    needs: [build, create-macos-universal]
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux
+            target: x86_64-unknown-linux-musl
+            runner: ubuntu-24.04
+          - platform: macos
```
xermicus: Looks like this is only run on a single macos platform, if so why is the universal binary needed after all? Is there a chance that the universal binary behaves different?

xermicus (approved):
Good stuff! This is looking much better now, thanks for taking on all that effort!

.github/workflows/reusable-build.yml
elle-j: Yeah this was previously never using the downloaded solc. I detected this after adding the caching of the builds. Before caching, the build step always ran, which invoked `make install-wasm` which invoked `make install-npm` which installed the `solc` npm package defined in [package.json](https://github.com/paritytech/revive/blob/90f4c070abdee086220b0b2e4127b4b0590ae8e3/js/emscripten/package.json#L5).

After adding caching, the build step didn't run on cache hit, so the sanity check failed since it couldn't find the npm package.

.github/workflows/reusable-build.yml
elle-j: I moved it before the sanity check so that the sanity check uses the same location as the upload artifact step uses (in short, so that it uses `require('./resolc-wasm32-unknown-emscripten/resolc.js')` instead of `require('./target/wasm32-unknown-emscripten/release/resolc.js')`).

.github/workflows/test-builds.yml
elle-j: Sounds good!

.github/workflows/test-builds.yml
elle-j: Ah good catch! Never added the `macos-15-intel` runner, but I think it'd be good to include. I'll add that.

PR #497, opened 2026-03-27
Build LLVM for Windows using MSVC-compatible Clang

elle-j:
## Description

Switches the Windows LLVM build from Visual Studio to `clang-cl` + `lld-link` (Clang's MSVC-compatible drivers) with Ninja.

This is in order to better match the Clang toolchain used on our Linux and macOS variants, while keeping the MSVC ABI.

It also speeds up the build time by at least 2x on Windows.


.github/workflows/release-llvm.yml
```diff
@@ -101,10 +104,21 @@ jobs:
...
+          arch: x64
+
       - name: Install LLVM Builder
         run: |
           cargo install --locked --force --path crates/llvm-builder
 
+      - name: Apply LLVM fix (temporary)
+        shell: bash
+        run: git -C llvm apply ../patches/llvm-190219.patch
+
```
xermicus: Can we please move this into the llvm-builder? So that it works when building everything locally too

xermicus (approved):
Nice findings thanks!

patches/llvm-190219.patch
xermicus, line 1: Don't want to scope creep but we could think about drive-bying @kvpanch [fix](https://github.com/llvm/llvm-project/pull/163679) too, seems simple enough

elle-j: Let's do it, I'm on it 👍

.github/workflows/release-llvm.yml
kvpanch: yea. Also, I'd suggest to also run `git reset HEAD` for `llvm` submodule, so  that running llvm-builder second time won't fail due to "previously applied patches"

elle-j: Yeah good idea! I ended up using `git apply --reverse --check` instead in the [PR](https://github.com/paritytech/revive/pull/500) where I've moved the patches to.

PR #500, opened 2026-04-08
Apply LLVM patches for deterministic operand order and correct peephole optimization for `minsize`

elle-j:
## Description

Applies LLVM patches that have not been backported to the major version used in our submodule:

* [LLVM PR #163679](https://github.com/llvm/llvm-project/pull/163679)
  * *"[RISCV] Allow large div peephole optimization for minsize"*
  * Fixes peephole optimization of large integers when `minsize` function attribute is set.
  * (Our revive codebase used a workaround that can be removed once this 500 PR is merged and we publish a new LLVM release.)
* [LLVM PR #190219](https://github.com/llvm/llvm-project/pull/190219)
  * *"[TargetLowering] Speculative fix for a non-determinism issue between different compilers"*
  * Operand ordering of `mul i256` were different depending on the target ABI when building LLVM, producing different RISC-V and PVM bytecode for our Windows vs macOS/linux builds.
  * The bytecode divergence (yet functionally equivalent) affected Solidity contracts containing a value [multiplied by itself](https://github.com/llvm/llvm-project/issues/190148) (e.g. via explicit squaring or exponentiation), manifesting primarily when using optimization level `-O0`.


crates/llvm-builder/src/utils.rs
```diff
@@ -10,12 +10,6 @@ use std::time::Duration;
 use anyhow::Context;
 use path_slash::PathBufExt;
 
-/// The LLVM host repository URL.
-pub const LLVM_HOST_SOURCE_URL: &str = "https://github.com/llvm/llvm-project";
-
-/// The LLVM host repository tag.
-pub const LLVM_HOST_SOURCE_TAG: &str = "llvmorg-18.1.8";
-
```
elle-j, line 18: Dead code.

.github/workflows/test.yml
```diff
@@ -35,6 +35,8 @@ jobs:
         uses: ./.github/actions/get-llvm
         with:
           target: x86_64-unknown-linux-gnu
+          # TODO: Remove (testing draft release)
+          version: "llvm-21.1.8-revive.0f80912"
```
elle-j: Will remove this.

.github/workflows/test.yml
```diff
@@ -35,6 +35,8 @@ jobs:
         uses: ./.github/actions/get-llvm
         with:
           target: x86_64-unknown-linux-gnu
+          # TODO: Remove (testing draft release)
+          version: "llvm-21.1.8-revive.5789e11"
```
elle-j: Will remove this.

crates/llvm-builder/src/utils.rs
kvpanch: good catch.

crates/llvm-builder/tests/common.rs
```diff
@@ -24,7 +24,7 @@ impl TestDir {
                 "submodule",
                 "add",
                 "-b",
-                "release/18.x",
+                "release/21.x",
```
kvpanch, line 27: hm.... I wonder how that even worked

kvpanch (approved):

crates/llvm-builder/tests/common.rs
elle-j: Seems like the tests using the `TestDir` use a fresh submodule with the tip of release/\<major\>.x.

crates/llvm-context/src/polkavm/context/mod.rs
xermicus, line 1: This should go into a follow-up; we first need to release a patched LLVM build

PR #505, opened 2026-04-17
Update LLVM from 21 to 22

kvpanch:


.github/actions/get-llvm/action.yml
```diff
@@ -88,22 +88,9 @@ runs:
...
-            }
-          } else {
-            // Find latest LLVM release (first in sorted list)
-            llvmRelease = llvmReleases[0];
-            if (llvmRelease) {
-              core.info(`Selected latest LLVM version: ${llvmRelease.tag_name}`);
-            }
+          let llvmRelease = allReleases.find(r => r.html_url && r.html_url.endsWith('/releases/tag/untagged-7b697020fcbbbca97a94'));
+          if (llvmRelease) {
+            core.info(`Selected pinned draft LLVM release: ${llvmRelease.html_url}`);
```
elle-j: This was likely a temporary change, this still uses an exact `untagged-`.

book/src/developer_guide/cross_compilation.md
elle-j, line 1: After changes to the `book`, also run `make book` locally and commit the updated `docs` directory 👍

.github/actions/get-llvm/action.yml
```diff
@@ -88,22 +88,9 @@ runs:
...
-            if (llvmRelease) {
-              core.info(`Selected LLVM release matching prefix '${versionPrefix}': ${llvmRelease.tag_name}`);
-            }
-          } else {
-            // Find latest LLVM release (first in sorted list)
-            llvmRelease = llvmReleases[0];
-            if (llvmRelease) {
-              core.info(`Selected latest LLVM version: ${llvmRelease.tag_name}`);
-            }
+          let llvmRelease = allReleases.find(r => r.html_url && r.html_url.endsWith('/releases/tag/untagged-7b697020fcbbbca97a94'));
```
xermicus: Should probably be rolled back?

xermicus (commented):
lgtm! could also update the `CLAUDE.md` file still referencing an older version

elle-j (approved):
Nice 🙂 After addressing comments 👍 

Also, does this update include at least the `22.1.3` release? That's where our patch [was applied](https://github.com/llvm/llvm-project/blob/e9846648fd6183ee6d8cbdb4502213fcf902a211/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp#L11241-L11247).

If both patches in [patches/llvm/](https://github.com/paritytech/revive/tree/main/patches/llvm) are included in the updated LLVM version used, then the corresponding `.patch` files could be removed perhaps in a follow-up. (The way our patches are applied, the only thing needed to apply or remove patches is to add or remove the `.patch` files.)

.github/actions/get-llvm/action.yml
kvpanch: yep. Just wanted to first confirm with you that everything else is fine.

kvpanch:
> Nice 🙂 After addressing comments 👍
> 
> Also, does this update include at least the `22.1.3` release? That's where our patch [was applied](https://github.com/llvm/llvm-project/blob/e9846648fd6183ee6d8cbdb4502213fcf902a211/llvm/lib/CodeGen/SelectionDAG/TargetLowering.cpp#L11241-L11247).
> 
> If both patches in [patches/llvm/](https://github.com/paritytech/revive/tree/main/patches/llvm) are included in the updated LLVM version used, then the corresponding `.patch` files could be removed perhaps in a follow-up. (The way our patches are applied, the only thing needed to apply or remove patches is to add or remove the `.patch` files.)

yes, that should include it. In next PR I will try to drop patches we have and will update LLVM commit

elle-j:
Just remembered, a CHANGELOG.md entry would be great as well

PR #509, opened 2026-04-23
update cargo dependencies

xermicus:
Fixes some security issue in a transitive dependency.


xermicus:
Dependabot was faster

PR #510, opened 2026-04-28
Update LLVM 22 to ca0698598261 and remove LLVM patches

kvpanch:


CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Supported `polkadot-sdk` rev: `2603`
```
elle-j, line 5: Saw that this entry was already part of the previous 1.1.0 release.

```suggestion
### Changed

```

elle-j (approved):

CHANGELOG.md
xermicus: Please do not remove this from the changelog. Its important for builders to see which version of the polkadot-sdk this compiler release will support.

elle-j: Ah it looked like it should only be added when it first changed, but can see in the CHANGELOG that the entry is kept for each release. Will update the suggestion to only add the `Changed` heading.

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
...
-- Updated `LLVM` from `21.1.8` to LLVM `22.1.3`
+- Updated `LLVM` from `21.1.8` to LLVM `22.1.4`
```
elle-j, line 9:
```suggestion

### Changed

- Updated `LLVM` from `21.1.8` to LLVM `22.1.4`
```

kvpanch: Do we really need to have explicit "Changed" in changelog ?

elle-j: From what I see in the CHANGELOG we're currently using `Added`, `Changed`, and `Fixed`.

PR #512, opened 2026-04-30
Support solc v0.8.35

xermicus:
- Support solc v0.8.35
- The new `erc7201` is comptime (already translated to a constant in Yul)

Didn't spot any relevant breakage or addition in the release notes so should be no-brainer.


crates/resolc/src/solc/mod.rs
elle-j, line 1: Regarding the new `--experimental`/`settings.experimental` mode toggling, could that be something we'd need to allow? Currently it looks like the only thing that would break users is if they have `pragma experimental solidity` in the source file, which would be a ["hard error"](https://www.soliditylang.org/blog/2026/04/29/solidity-0.8.35-release-announcement#the---experimental-flag) without the `experimental` flag.

xermicus: Good point! But its futile: In resolc we always need `--via-ir` but `--via-ir` and `--experimental` are mutually exclusive.

PR #522, opened 2026-05-21
Add reference page for newyork IR

elle-j:
Adds a per-operation reference for the newyork IR to the compiler book for easily being able to view printed syntax, operand and result types, examples, etc. for any operation. Using easily readable syntax notation. The new page is partly inspired by [MLIR's docs](https://mlir.llvm.org/docs/Dialects/SCFDialect/).

<details>
  <summary>Example entry</summary>

### `and`

(`Expression::Binary` with `BinaryOperation::And`)

#### Description

Bitwise AND. The common idiom for type narrowing: a constant mask on the right lets forward analysis pick up a tight result width.

#### Syntax

```text
and($lhs[: <type>], $rhs[: <type>])
```

#### Example

```text
let v2 := and(v0, v1)
let v3 := and(v0, 0xff)     // type inference narrows result to i8
```

#### Operands

| Name | Type | Notes |
|---|---|---|
| `lhs` | `i256` | — |
| `rhs` | `i256` | — |

#### Result and purity

| Result | Purity |
|---|---|
| `min(width(lhs), width(rhs))` — AND can only clear bits, so the result fits in the narrower operand | Pure |

#### Annotations

None.

</details>


xermicus (approved):
Thanks

Issue #216, opened 2025-02-18
Code coverage


xermicus:
Generate code coverage reports (in `README.md` and as PRs comments). Start with the `integration` tests, (once ready, the E2E validation testing suite should be used to paint a more complete picture). Requires support in the LLVM build.

PR #527, opened 2026-05-26
Fix optimization specific SREM edge case

### LINKED ISSUE #524: Miscompile in `__revive_signed_remainder`: `srem(INT256_MIN, -1)` UB
`SignedRemainder` lowers Solidity signed `%` (EVM `SMOD`) to a bare LLVM `srem` via [`build_int_signed_rem`](https://github.com/paritytech/revive/blob/7fe8d41/crates/llvm-context/src/polkavm/context/function/runtime/arithmetics.rs#L185), guarded only against a zero divisor by [`wrapped_division`](https://github.com/paritytech/revive/blob/7fe8d41/crates/llvm-context/src/polkavm/context/function/runtime/arithmetics.rs#L208). For divisor `-1` it emits `srem x, -1`, which is UB in LLVM when `x == INT256_MIN` and constant-folds to poison.

MRE:
```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.20;
contract C {
    function f(int256 x) external pure returns (int256) {
        // INT_MIN % -1 == 0, so this is the identity `return x;`
        unchecked { return (type(int256).min % ((int256(-1) << 58) >> 58)) ^ x; }
    }
}
```

Reproduce: compile with `resolc -O3 --bin mre.sol` and call `f(5)` on PolkaVM/pallet-revive — it returns `0`. The same source compiled with `solc --bin-runtime --optimize` and run on go-ethereum `evm` returns `5`.

```
EVM:     f(5) = 0x...05   (correct)
PolkaVM: f(5) = 0x...00   (whole function collapsed to 0)
```

Expected behavior: `f(x) == x` for all `x` (EVM `SMOD(INT_MIN, -1) = 0`).

Notes:
- Only signed `%` (`srem`); `/` (`sdiv`) is guarded and correct.
- Runtime `x % y` with `(INT_MIN, -1)` is correct — the bare `srem` lowers to RISC-V `rem`, which defines `INT_MIN rem -1 = 0`; the bug is only LLVM's constant-fold of `srem(constINT_MIN, const-1)`, so the dividend must also be compile-time constant.
- The divisor must be a `-1` that solc does not pre-fold to a literal (here `(int256(-1) << 58) >> 58`); writing `int256(-1)` directly makes solc fold `INT_MIN % -1 → 0` itself and the bug disappears.
- Silent: no diagnostic; PolkaVM bytecode returns the wrong constant.

Git commit: 7fe8d41 (revive)
Build: resolc release, -O3; solc 0.8.35-develop; go-ethereum evm 1.17.2

xermicus:
Existing test coverage didn't catch a signed remainder overflow semantics edge case behavior where at least one the operators is being constant folded by LLVM instead of solc. This PR fixes the SREM translation bug and adds missing test coverage for division ops with constants.

Closes #524

### DIFF
--- a/crates/integration/src/tests.rs
+++ b/crates/integration/src/tests.rs
@@ -7,7 +7,7 @@ use revive_runner::*;
 use SpecsAction::*;
 
 use crate::cases::Contract;
-use crate::cases::DivisionArithmeticsConst;
+use crate::cases::{DivConst, ModConst, SdivConst, SmodConst};
 
 /// Parameters:
 /// - The function name of the test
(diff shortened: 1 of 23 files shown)


kvpanch (approved):
maybe we need to add fuzzer similar to PVM ?

xermicus:
> maybe we need to add fuzzer similar to PVM ?

Yes we do. [Its on the back of my mind since a long while.](https://github.com/paritytech/revive/issues/216) This is also the reason why I the LLVM builder supports building with coverage.

crates/integration/src/tests.rs
```diff
@@ -268,6 +270,500 @@ fn signed_remainder() {
...
+    let pair_count = 17;
+    for i in 0..pair_count {
+        push_call(&mut actions, TestAddress::Instantiated(0), yul_which_calldata(i));
+    }
+    run_differential(actions);
+}
+
+/// Surfaces the `smod(INT_MIN, -1)` LLVM-UB const-fold bug from
+/// paritytech/revive#524. See `SmodIntMinNegOneBug.yul` for why this specific
+/// fixture is shaped the way it is. Expected to FAIL until the bug is fixed.
```
elle-j, line 686:
```suggestion
/// fixture is shaped the way it is.
```

elle-j (approved):
LGTM, could you add a CHANGELOG entry briefly describing this fix?

PR #529, opened 2026-05-26
Code coverage

kvpanch:
Adds `make coverage` (cargo-llvm-cov over the full workspace including revive-llvm-builder), a new "Code coverage" book chapter that briefly explains how to run it, and a PR workflow that runs coverage and comments on the PR (it updates the same comment if it already exists) when the `measure code coverage` label is applied (see e.g. [this comment](https://github.com/paritytech/revive/pull/529#issuecomment-4966136441)).

Building instrumented LLVM and generating an LLVM C++ coverage report is also possible, see accompanying [PR description #575](https://github.com/paritytech/revive/pull/575) for more details.


Makefile
```diff
@@ -85,10 +89,60 @@ test-llvm-builder:
 	@echo "warning: the llvm-builder tests will take many hours"
 	cargo test --package revive-llvm-builder -- --test-threads=1
 
-test-book:
-	cargo install mdbook --version 0.5.1 --locked
+MDBOOK_VERSION = 0.5.1
+
+install-mdbook:
+	@mdbook --version 2>/dev/null | grep -qx "mdbook v$(MDBOOK_VERSION)" \
```
xermicus: Is this necessary? The cargo install is essentially a no-op if the installed version is already present

Makefile
```diff
@@ -85,10 +89,60 @@ test-llvm-builder:
...
+	@command -v cargo-llvm-cov >/dev/null 2>&1 || cargo install cargo-llvm-cov --locked
+	@rustup component add llvm-tools-preview >/dev/null 2>&1 || true
+
+test-book: install-mdbook
 	mdbook test book
 
+# Workspace-wide coverage (including revive-llvm-builder). Stages HTML under
+# book/src/coverage/; mdbook copies it to docs/coverage/ during build. The
+# chapter is stamped in place — revert with `make coverage-reset`.
+# Several hours expected: llvm-builder tests run single-threaded.
```
xermicus: Hmm so this `coverage` command shouldn't execute LLVM builder tests? I think it ultimately should collectc coverage data from the unit tests of resolc crates, including the integration tests. Once it works well enough we maybe also want have coverage reports the from retester corpora.

xermicus (commented):
Quick first round

Makefile
kvpanch: I wanted to cover LLVM Builder too as it would help to catch dead code, but now I do think that's an overkill.

xermicus: Yeah I don't think its necessary. Instead it should link in an LLVM build with coverage instrumentation too, so we get the full picture.

Makefile
kvpanch: will drop it.

revive-differential-tests
elle-j, line 1: This probably got accidentally pushed. You can locally checkout the same `revive-differential-tests` and `revive-differential-tests/resolc-compiler-tests` commits that `main` references, and push that.

kvpanch: oops. Yes, I didn't want to

xermicus (commented):
Would be nice to get these fixes in. @kvpanch can you please resolve the merge conflicts?

revive-differential-tests
elle-j: @kvpanch , please also make sure `revive-differential-tests` is not listed in these changed files 👍

xermicus:
Pushed to the branch since @kvpanch is OOO.

Eventually I scrapped the book approach. It's too brittle. I vote: Merge this now as is with just the fixes, and think a bit more about where the coverage report should live. Or if we should overhaul the approach of checking in the build `docs/` altogether (though I like the simple and stupid approach).

elle-j (commented):
Let's preferably hold off with this merge atm, found some security issues and other fixes that I'll address as a PR targeting this branch.

elle-j (approved):
Will be great to use this! 🥳

PR #544, opened 2026-06-09
Fix Wasm build cache collision

elle-j:
## Description

Before the CI refactor, the Wasm builds initiated by `release.yml` and `test-wasm.yml` ran in parallel, building duplicate builds. After the CI refactor, those workflows shared the build cache key to instead reuse the build. In both cases, the resolc Wasm URI baked into `resolc_web.js` depends on whether it is a release or not.

If the Wasm build step initiated by `release.yml` started *before* `test-wasm.yml`, `test-wasm.yml` would use an incorrect cache due to the invalid resolc Wasm URI (seen by [this failed run](https://github.com/paritytech/revive/actions/runs/27178950108/job/80235255116?pr=529)).

This PR updates the workflow input `is_release` to only pertain to ref names starting with `v` (which is what the `create-release` step guards on), and updates the cache key and concurrency group to include `is_release`, solving the cache collision. (Thus, the workflows needing a Wasm build for non-v-releases will still reuse the now-correct cache.)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,7 +67,7 @@ jobs:
     needs: [check-version-changed]
     uses: ./.github/workflows/reusable-build.yml
     with:
-      is_release: true
+      is_release: ${{ startsWith(github.ref_name, 'v') }}
       retention_days: 1
 
   create-release:
--- a/.github/workflows/reusable-build-wasm.yml
+++ b/.github/workflows/reusable-build-wasm.yml
@@ -32,7 +32,7 @@ jobs:
       resolc_web_js_sha: ${{ steps.upload-web-js.outputs.artifact-digest }}
     concurrency:
       # Use `reusable-build` and not `${{ github.workflow }}` here so that callers share the same concurrency group.
-      group: reusable-build-wasm32-unknown-emscripten-${{ github.sha }}
+      group: reusable-build-wasm32-unknown-emscripten-is-release-${{ inputs.is_release }}-${{ github.sha }}
       cancel-in-progress: false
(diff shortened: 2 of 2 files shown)


kvpanch:
> If the Wasm build step initiated by release.yml started before test-wasm.yml, test-wasm.yml would use an incorrect cache due to the invalid resolc Wasm URI 

just for my understanding: can this/does it make sense to start execution in order ? If this keep happening, won't two jobs build same thing twice ?

elle-j:
> can this/does it make sense to start execution in order ? If this keep happening, won't two jobs build same thing twice ?

Before the whole CI refactor, Wasm was built twice (in parallel). It doesn't build twice any more, the one that starts first builds and caches it, the second waits until the first finishes and uses its cache. Both `release.yml` and `test-wasm.yml` are top-level and either one could be faster with initiating the Wasm build step (usually test-wasm.yml has been faster), but it shouldn't matter which one starts first.

kvpanch (approved):

kvpanch:
> > can this/does it make sense to start execution in order ? If this keep happening, won't two jobs build same thing twice ?
> 
> Before the whole CI refactor, Wasm was built twice (in parallel). It doesn't build twice any more, the one that starts first builds and caches it, the second waits until the first finishes and uses its cache. Both `release.yml` and `test-wasm.yml` are top-level and either one could be faster with initiating the Wasm build step (usually test-wasm.yml has been faster), but it shouldn't matter which one starts first.

thanks for clarification. Basically, cache here is a dependency between jobs

xermicus (approved):

PR #545, opened 2026-06-16
bump cargo dependencies

xermicus:
- Update PVM to fix a downstream bug.
- Update the polkadot-sdk to get rid of most retester exceptions.


Cargo.toml
```diff
@@ -74,7 +74,7 @@ normpath = "1.5"
 # polkadot-sdk and friends
 codec = { version = "3.7.5", default-features = false, package = "parity-scale-codec" }
 scale-info = { version = "2.11.6", default-features = false }
-polkadot-sdk = { version = "=2603.0.0" }
+polkadot-sdk = { version = "=2604.2.0" }
```
elle-j, line 77: Can you update `.github/workflows/differential-tests.yml` as well:
https://github.com/paritytech/revive/blob/11e2ea1be4ee9d2efd30121b6b6f67397275219c/.github/workflows/differential-tests.yml#L139-L143

```diff
     env:
       # Commit SHA that produced the polkadot-sdk crate pinned in Cargo.toml,
       # obtained from the crate's .cargo_vcs_info.json via:
-      # curl -sL https://crates.io/api/v1/crates/polkadot-sdk/<VERSION>/download | tar xzO '*/.cargo_vcs_info.json'
-      # Current version: 2603.0.0
-      POLKADOT_SDK_REF: "c56b086a3cfcb3a48b75fd1ef9c77ad9b7408b95"
+      # curl -sL -A "user-agent" https://crates.io/api/v1/crates/polkadot-sdk/<VERSION>/download | tar xzO polkadot-sdk-<VERSION>/.cargo_vcs_info.json
+      # Current version: 2604.2.0
+      POLKADOT_SDK_REF: "54095b130013902aaaa33a69dbbf980144764afb"
```

CHANGELOG.md
elle-j, line 1: Saw that we've [previously](https://github.com/paritytech/revive/blob/11e2ea1be4ee9d2efd30121b6b6f67397275219c/CHANGELOG.md?plain=1#L390) also included linker fixes in the changelog. Could be useful for this bump as well :)

elle-j (approved):
LGTM after minor updates 👍

CHANGELOG.md
xermicus: This was an actual fix on our end and not just a dependency update. Lets not mention downstream fixes in our changelog. That would seriously clutter it.

Cargo.toml
xermicus: Good catch thanks!

PR #555, opened 2026-06-29
Extend differential tests with newyork optimizer

### LINKED ISSUE #553: Support compiling differential tests via newyork IR
Our differential tests (both "compile + compare hashes" and "running e2e tests") should also exercise the new newyork IR pipeline, starting with `Oz`. This requires updates in retester.

elle-j:
## Description

Extends our differential tests such that the compile+compare bytecode hashes and the E2E tests also exercise the newyork IR pipeline.

The requested compiler modes that E2E test cases run with are still controlled by the metadata in the corpus, but the related PR listed below also enables support for an allow-list to optionally narrow the set of modes run, via an input to the GH action used.

The changes to the workflow in this PR allows all newyork retester compiler modes (all LLVM opt levels + solc optimizer on/off) to run for the E2E tests since they still complete fast.

## Related

The PRs implemented as prerequisites alongside this PR are:
* `revive-differential-tests`
  * [Support optimized newyork IR compiler pipeline](https://github.com/paritytech/revive-differential-tests/pull/308)
* `resolc-compiler-tests`
  * [Add newyork IR pipeline counterparts for explicitly-listed Yul modes](https://github.com/paritytech/resolc-compiler-tests/pull/15)

## Resolved issues

Closes #553

### DIFF
--- a/.github/workflows/differential-tests.yml
+++ b/.github/workflows/differential-tests.yml
@@ -6,9 +6,10 @@ on:
 env:
   IS_PUSH_TO_MAIN: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
   # The exact compiler modes to use for compiling and comparing hashes by retester.
-  COMPILE_MODES: "Y M0 S+, Y M3 S+, Y Mz S+, NY Mz S+"
+  COMPILE_MODES: "NY M0 S+, NY M3 S+, NY Mz S+, NY M0 S-, NY M3 S-, NY Mz S-"
   # The compiler modes allowed to run during e2e test execution by retester.
-  E2E_ALLOWED_MODES: "Y, NY Mz S+, NY Mz S-"
+  # E.g. this can be updated to allow only `"Y, NY Mz S+"` (all Yul modes + newyork with -Oz and solc optimizer enabled).
+  E2E_ALLOWED_MODES: "Y, NY"
   SOLC_VERSION: "0.8.35"
 
 jobs:
--- a/.github/workflows/differential-tests.yml
+++ b/.github/workflows/differential-tests.yml
@@ -6,7 +6,11 @@ on:
 env:
   IS_PUSH_TO_MAIN: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
   # The exact compiler modes to use for compiling and comparing hashes by retester.
-  COMPILE_MODES: "Y M0 S+, Y M3 S+, Y Mz S+, NY M3 S+, NY Mz S+"
+  # Runs triggered by pushes to main exercise more modes.
+  COMPILE_MODES: >-
+    ${{ (github.event_name == 'push' && github.ref == 'refs/heads/main')
+    && 'Y M0 S+, Y M3 S+, Y Mz S+, NY M0 S+, NY M3 S+, NY Mz S+'
+    || 'Y M0 S+, Y M3 S+, Y Mz S+, NY Mz S+' }}
   # The compiler modes allowed to run during e2e test execution by retester.
   # E.g. this can be updated to allow only `"Y, NY Mz S+"` (all Yul modes + newyork with -Oz and solc optimizer enabled).
   E2E_ALLOWED_MODES: "Y, NY"
--- a/.github/workflows/differential-tests.yml
+++ b/.github/workflows/differential-tests.yml
@@ -6,7 +6,7 @@ on:
 env:
   IS_PUSH_TO_MAIN: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
   # The exact compiler modes to use for compiling and comparing hashes by retester.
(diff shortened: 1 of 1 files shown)


.github/workflows/differential-tests.yml
```diff
@@ -5,8 +5,11 @@ on:
 
 env:
   IS_PUSH_TO_MAIN: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
-  # The compiler modes to use for compilation by retester.
-  MODES: "Y M0 S+, Y M3 S+, Y Mz S+"
+  # The exact compiler modes to use for compiling and comparing hashes by retester.
+  COMPILE_MODES: "Y M0 S+, Y M3 S+, Y Mz S+, NY M3 S+, NY Mz S+"
+  # The compiler modes allowed to run during e2e test execution by retester.
+  # E.g. this can be updated to allow only `"Y, NY Mz S+"` (all Yul modes + newyork with -Oz and solc optimizer enabled).
+  E2E_ALLOWED_MODES: "Y, NY"
```
elle-j, line 16: The `E2E_ALLOWED_MODES` allow all requested modes that use Yul and newyork IR pipelines, which is still fast so it's good to keep this. The `COMPILE_MODES` currently uses a subset.

A bytecode mismatch was detected on Windows for 1 contract using `NY Mz S-` (newyork, `-Oz`, solc optimizer *disabled*) which I'm currently investigating. This PR will exclude that mode from the comparison so that it can be merged with the current settings, and I'll continue investigating the mismatch in the mean time.

.github/workflows/differential-tests.yml
```diff
@@ -5,8 +5,11 @@ on:
 
 env:
   IS_PUSH_TO_MAIN: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
-  # The compiler modes to use for compilation by retester.
-  MODES: "Y M0 S+, Y M3 S+, Y Mz S+"
+  # The exact compiler modes to use for compiling and comparing hashes by retester.
+  COMPILE_MODES: "Y M0 S+, Y M3 S+, Y Mz S+, NY M3 S+, NY Mz S+"
```
kvpanch: Why not test `NY M0 S+` ?

elle-j: We first discussed initially adding newyork running only with `Oz` (`Mz` in retester's representation) due to CI duration (having all 3 counterparts would double the compilation time). I think for now we could use these for PRs:
* `Y M0 S+`
* `Y M3 S+`
* `Y Mz S+`
* `NY Mz S+`

But for pushes to `main` we could *additionally* run:
* `NY M0 S+`
* `NY M3 S+`

Since macOS intel is extra slow (and runs only on `main`) having it also run these extra will take time, but if we experience that as a problem on `main` we could later look into separating `Y` and `NY` compilations onto different runners.

I know you mentioned perhaps having a cron job, that could probably be something we could do in addition to the above if we'd like to test `S-` as well. Would then schedule it to run during the night. Running with `S-` would be beneficial with `NY` since we currently have a bytecode mismatch with that combination (for 1 contract) that I'll investigate.

P.S. The e2e differential tests *are* running with *all* optimization combinations (LLVM + solc optimizer) with both Yul and newyork pipelines.

kvpanch (approved):

.github/workflows/differential-tests.yml
```diff
@@ -5,8 +5,15 @@ on:
 
 env:
   IS_PUSH_TO_MAIN: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
-  # The compiler modes to use for compilation by retester.
-  MODES: "Y M0 S+, Y M3 S+, Y Mz S+"
+  # The exact compiler modes to use for compiling and comparing hashes by retester.
+  # Runs triggered by pushes to main exercise more modes.
```
xermicus, line 9: Nice, I like this idea.

xermicus (approved):

.github/workflows/differential-tests.yml
xermicus: An optimization for testing reproducible builds: For NY skip machine code generation with LLVM entirely and compare the generated IR. Not sure if this works on byte-level (i.e. if it includes path), otherwise would need to parse it back and compare the datastructure.

NY is before LLVM, so it follows that if NY is reproducible, then the input to LLVM is reproducible.

For execution tests it could for example still be that NY generates buggy code but -Oz or -O3 optimizes it away, so I think there we have to exercise all combinations.

elle-j: For reproducible builds, I think we should still compare LLVM-generated bytecode given the bugs we've previously found in LLVM's instruction selection pass, which generated semantically equivalent bytecode (thus e2e tests do not catch this) but not identical bytecode (differed on Windows).

The e2e execution tests do run all combinations 👍

xermicus: Yes we would still need to compare all LLVM opt levels, but not for NY. Either ways its fine and would be just a small optimization.

PR #557, opened 2026-06-30
Restore resolc long version onto standard JSON and combined JSON output

elle-j:
## Description

The resolc long version stopped being populated on the standard JSON and combined JSON output. This PR restores these fields and populates them.


xermicus (approved):
Good catch thanks!

kvpanch (approved):

PR #566, opened 2026-07-07
Fix/ny unaligned store fmp oog

xermicus:
Bug found by @jubnzv:

`--newyork`: an unaligned `mstore` corrupting the free-memory-pointer word let a later store through the corrupted pointer silently succeed where EVM runs out of gas.


crates/newyork/src/heap_opt.rs
```diff
@@ -184,6 +191,55 @@ impl Default for OffsetInfo {
...
+///
+/// A sequence can be left by falling through to the next statement, but also by edges that
+/// skip intervening statements without terminating the frame: `break` (skips the loop `post`),
+/// `continue` (jumps to the loop `post`), and `leave` (returns to the caller). Each such edge
+/// carries its own corruption state, which the responsible enclosing construct — the `For`
+/// handler for `break`/`continue`, the per-function summary for `leave` — merges at the edge's
+/// actual target. Collapsing them into one state would let a path that re-establishes the
+/// pointer mask a corrupted path that skips it.
+#[derive(Clone, Copy, Default)]
+struct CorruptionExit {
```
elle-j: Since this is isolated specifically to fmp corruption, can we rename this to `FmpCorruptionExit`?

crates/newyork/src/heap_opt.rs
```diff
@@ -184,6 +191,55 @@ impl Default for OffsetInfo {
...
+    fn absorb_jump_exits(&mut self, nested: CorruptionExit) {
+        self.break_corrupt |= nested.break_corrupt;
+        self.continue_corrupt |= nested.continue_corrupt;
+        self.leave_corrupt |= nested.leave_corrupt;
+    }
+}
+
+/// One pass over a loop's per-iteration sequence for
+/// [`HeapAnalysis::scan_loop_fmp_corruption`].
+struct LoopIterationScan {
```
elle-j: This name also seems quite vague in relation to what it's used for. Perhaps name it something like `LoopIterationFmpCorruption`. (If so, along with `fn scan_loop_iteration()` --> `fn scan_loop_iteration_fmp_corruption()` which is more consistent with other functions such as e.g. `fn scan_expression_fmp_corruption()`.)

crates/newyork/src/heap_opt.rs
```diff
@@ -617,6 +676,284 @@ impl HeapAnalysis {
...
+                    ..
+                } => {
+                    corrupt = self.scan_loop_fmp_corruption(
+                        condition_statements,
+                        &body.statements,
+                        &post.statements,
+                        corrupt,
+                        &mut exit,
+                    );
+                }
```
elle-j, line 885: This is missing the `condition` expression. So the later loop iteration scan also can call `scan_expression_fmp_corruption()`.

Update: Similarly, I also noticed that the `HeapAnalysis::analyze_statement()`'s `For` match arm is also missing the `condition` and running `self.analyze_expression_side_effects(condition)`.

crates/newyork/src/mem_opt.rs
```diff
@@ -1130,9 +1146,9 @@ impl FmpPropagation {
                     mut post,
                     outputs,
                 } => {
-                    let loop_modifies_fmp = self.region_modifies_fmp(&body.statements)
-                        || self.region_modifies_fmp(&condition_statements)
-                        || self.region_modifies_fmp(&post.statements);
+                    let loop_modifies_fmp = self.region_modifies_fmp(&body.statements, &constants)
+                        || self.region_modifies_fmp(&condition_statements, &constants)
+                        || self.region_modifies_fmp(&post.statements, &constants);
```
elle-j: `fn region_modifies_fmp()` currently also checks whether any of the expression statements are `Expression::Call`s, and then checks `self.fmp_writers.contains(..)`. But this block does not check this for the for-`condition`.

If `fn region_modifies_fmp()` also accepted and handled e.g. an `&Expression` or `&[Expression]`, this block could additionally pass the `condition`, or do the `self.fmp_writers.contains(..)` check here.

crates/newyork/src/mem_opt.rs
```diff
@@ -1266,48 +1282,90 @@ impl FmpPropagation {
...
+    /// pointer), a statically-resolved `mstore8` into the word, or an `ExternalCall`/`Create`.
+    ///
+    /// Offsets are resolved against `constants` — the caller's resolution scope — extended with
+    /// literal bindings encountered during the (program-order) walk, mirroring the straight-line
+    /// resolution in `propagate_statements`. An offset bound outside the passed scope resolves to
+    /// `None` and falls back to the region tag, exactly as straight-line code does. A dynamic
+    /// (unresolvable) store or `mstore8` that could wrap onto the FMP word is the same deliberate,
+    /// solc-unreachable gap as in straight-line `mstore8` handling (see the
+    /// `fmp_could_be_unbounded` field docs in `heap_opt`).
+    fn statements_store_to_fmp_word(
```
elle-j, line 1329: This is missing handling of e.g. the `*Copy` statements. (The doc here also says *"..mirroring the straight-line resolution in `propagate_statements`"*, and `propagate_statements` do include the `*Copy` statements.)

crates/newyork/src/heap_opt.rs
elle-j: Looks like this doc comment should be updated. E.g. it currently says *"..distinguishing observed from unobserved corruption needs a control-flow dataflow this object-wide flag does not model"*.

crates/newyork/src/heap_opt.rs
xermicus: Yeah good point

crates/newyork/src/heap_opt.rs
xermicus: I agree, can name it more explicitly

crates/newyork/src/mem_opt.rs
xermicus: Good catch

crates/newyork/src/heap_opt.rs
xermicus: Yes the LLM models (and me too haha) chronically forget to keep docs in sync

crates/newyork/src/heap_opt.rs
xermicus: Good catch, I think you are right!

crates/newyork/src/heap_opt.rs
```diff
@@ -347,9 +407,11 @@ impl HeapAnalysis {
             Statement::For {
                 initial_values,
                 loop_variables,
+                condition,
                 outputs,
                 ..
             } => {
+                self.analyze_expression_side_effects(condition);
```
elle-j: Haven't dug into this more but one thought arising here is whether `self.analyze_expression_side_effects()` is correct to have *before* or *after* the iteration of the `initial_values`. It may not matter in this case, but could be worth quickly checking this.

elle-j (approved):
Nice, LGTM!

crates/newyork/src/heap_opt.rs
xermicus: Makes sense to me to have it before

PR #567, opened 2026-07-07
Fix/ny fn native fmp oog

xermicus:
Bug found by @jubnzv:

`--newyork`: a full-width value stored to the free-memory-pointer slot (e.g. a function parameter `>= 2^64`) was narrowed to 64 bits, inserting a checked truncation that consumed all gas.


crates/newyork/src/type_inference.rs
elle-j, line 1066: Side note: Perhaps this `For` arm should also use the `condition` and do `self.collect_uses_expression(condition)`.

elle-j (approved):

crates/newyork/src/type_inference.rs
xermicus: Nice catch thanks!

PR #569, opened 2026-07-08
Fix rate-limited cloning of LLVM submodule

elle-j:
## Description

We started seeing rate-limit `429` errors when cloning LLVM (as seen in e.g. [this failed run](https://github.com/paritytech/revive/actions/runs/28924583524/job/85809154518) and [this failed run](https://github.com/paritytech/revive/actions/runs/28926259175/job/85821336465?pr=564)). `llvm/llvm-project` now seems to be heavily rate-limited by GitHub. (Authenticating the clone did not seem to have much effect.)

Even though our resolc build steps only do a treeless clone, all build steps still did the same clone.

### Changes

* I've instead introduced a `.llvm-version` file that the `get-llvm` action reads from, eliminating the network request all together.
  * `.llvm-version` is meant to be kept in sync with the LLVM submodule version we use. This is enforced by:
  * Introduced a `check-llvm-version.yml` workflow
    * Runs once to check that the recorded version in `.llvm-version` coincides with the version detected from the submodule.
    * The submodule version detection *does* perform a network request as before, but this is only run once now (when relevant files change), rather than for all our resolc build jobs.
* Since the differential test jobs require checking out our revive-differential-tests (rdt) submodule, those checkout steps cloned not only rdt, but also LLVM. I've updated this to not clone LLVM at all.


xermicus (commented):
LGTM! I like the `.llvm-version` approach, this was starting to get a bit messy anyways.

xermicus:
Can we limit running 1 or maybe 2 PRs CI pipelines at most, and others have to queue up and wait? This should fix the problem in more depth and also prevent approved contributors from (intentially or not) wasting a lot of CI resources. Most PRs will need a merge main and re-run the pipeline anyways if multiple PRs are on the way.

elle-j:
> Can we limit running 1 or maybe 2 PRs CI pipelines at most, and others have to queue up and wait?

Can definitely look into that! Yeah I think at least 2 PRs at the same time would be useful in order to get feedback faster from CI runs, rather than 1.

PR #571, opened 2026-07-10
new clippies

xermicus:


crates/newyork/src/simplify.rs
```diff
@@ -970,10 +970,9 @@ impl Simplifier {
                 let (exponent, value) =
                     if let Some(exponent) = rhs_value.as_ref().and_then(log2_exact) {
                         (exponent, lhs)
-                    } else if let Some(exponent) = lhs_value.as_ref().and_then(log2_exact) {
-                        (exponent, rhs)
                     } else {
-                        return None;
+                        let exponent = lhs_value.as_ref().and_then(log2_exact)?;
```
xermicus, line 974: Not that this is any better lol. But whatever

PR #572, opened 2026-07-10
support solc v0.8.36

xermicus:


elle-j (approved):
Nice timing 🙂 Could you also update the one [in foundry](https://github.com/paritytech/revive/blob/e7c761c0688fd08eb11186b56295b52551f3bde4/tooling-projects/foundry/erc20/foundry.toml#L5) and [in hardhat](https://github.com/paritytech/revive/blob/e7c761c0688fd08eb11186b56295b52551f3bde4/tooling-projects/hardhat/erc20/hardhat.config.js#L17) 👍

xermicus:
Nice catch thanks, they are new and I didn't think about it

PR #564, opened 2026-07-07
Fix/ny dynamic calldatacopy fmp

xermicus:
Bug found by @jubnzv:

`--newyork`: a `calldatacopy`/`codecopy`/`returndatacopy` to a dynamic destination that could overwrite the free-memory-pointer word truncated a subsequent `mload(0x40)` to `0`.

PR #561, opened 2026-07-06
Fix NY codegen wth folded guard + inlined for-loop

xermicus:
Fixes a ICE wth folded guard in inlined for-loop in NY.

Closes #560

PR #573, opened 2026-07-10
release resolc v1.4.0

xermicus:
resolc v1.4.0

Supported `polkadot-sdk` rev: `2604.2.0`

### Added
- Support for solc v0.8.36 and the "Amsterdam" EVM version.
### Fixed
- Restored the top-level `revive_version` field in `--standard-json` output and the `resolc_version` field in `--combined-json` output, both of which stopped being populated in `v0.4.0`. [#557](https://github.com/paritytech/revive/pull/557)
- ICE triggered by `--disable-solc-optimizer --newyork` where inlined loops could emit an unterminated basic block in the inlined entrypoint. [#561](https://github.com/paritytech/revive/pull/561)
- `--newyork`: dead-store elimination could drop a store whose memory expansion an intervening `msize()` observes, so `msize()` read a stale (unexpanded) value. [#565](https://github.com/paritytech/revive/pull/565)
- `--newyork`: dead-store elimination could drop a store that a later overlapping unaligned `mload` reads, returning zeroed memory instead of the stored bytes. [#563](https://github.com/paritytech/revive/pull/563)
- `--newyork`: an unaligned full-word `mload`/`mstore` overlapping a word kept native (little-endian) by the heap optimization returned byte-swapped bytes. [#562](https://github.com/paritytech/revive/pull/562)
- `--newyork`: a `calldatacopy`/`codecopy`/`returndatacopy` to a dynamic destination that could overwrite the free-memory-pointer word truncated a subsequent `mload(0x40)` to `0`. [#564](https://github.com/paritytech/revive/pull/564)
- `--newyork`: a full-width value stored to the free-memory-pointer slot (e.g. a function parameter `>= 2^64`) was narrowed to 64 bits, inserting a checked truncation that consumed all gas. [#567](https://github.com/paritytech/revive/pull/567)
- `--newyork`: an unaligned `mstore` corrupting the free-memory-pointer word let a later store through the corrupted pointer silently succeed where EVM runs out of gas. [#566](https://github.com/paritytech/revive/pull/566)
- Fixed a bytecode divergence (yet functionally equivalent) between the resolc Linux/macOS builds and the Windows build for contracts with loops indexing array elements at constant offsets from a `uint8` loop variable (e.g. `array[i]`/`array[i + 1]`), compiled with `--newyork` at `-Oz` with the solc optimizer disabled. [#570](https://github.com/paritytech/revive/pull/570)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## v1.4.0
 
 Supported `polkadot-sdk` rev: `2604.2.0`
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9427,7 +9427,7 @@ dependencies = [
 
 [[package]]
 name = "resolc"
-version = "1.3.0"
+version = "1.4.0"
 dependencies = [
  "anyhow",
  "clap",
@@ -9481,7 +9481,7 @@ dependencies = [
 
 [[package]]
 name = "revive-common"
-version = "1.1.0"
+version = "1.2.0"
 dependencies = [
  "anyhow",
  "hex",
@@ -9506,7 +9506,7 @@ dependencies = [
 
 [[package]]
 name = "revive-integration"
-version = "1.3.0"
+version = "1.4.0"
 dependencies = [
  "alloy-primitives",
  "alloy-sol-types",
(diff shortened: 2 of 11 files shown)


elle-j (approved):
🥳

Issue #581, opened 2026-07-27
ICE: integer overflow in heap_opt::taint_range with --newyork on valid Yul


Esorat:
`resolc --yul --newyork --bin` panics with "attempt to add with overflow" at `crates/newyork/src/optimizer/heap_opt.rs:724` on valid Yul input. The panic occurs in the `taint_range` function when processing `mstore(add(mul(MAX_U256, 1), 0x41), v)` — the multiplication+addition overflows `usize` in range arithmetic.

**MRE:**
```yul
object "C" {
  code { datacopy(0, dataoffset("C_deployed"), datasize("C_deployed")) return(0, datasize("C_deployed")) }
  object "C_deployed" {
    code {
      mstore(0x40, 0x80)
      mstore(add(mul(0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff, 1), 0x41),
             0x000000000000000000000000000000000000000000000000ffffffffffffffff)
      let fmp := mload(0x40)
      let f2 := mload(0x40)
      mstore(f2, 0xC0FFEE)
      let rb := mload(f2)
      mstore(0, fmp)
      mstore(32, rb)
      return(0, 64)
    }
  }
}
```

**To reproduce:**
```
$ resolc --yul --bin -O 3            # OK
$ resolc --yul --newyork --bin -O 3  # PANIC at heap_opt.rs:724
```

The Yul is valid — `mul(MAX_U256, 1)` produces `MAX_U256`, and `add(MAX_U256, 0x41)` is valid 256-bit arithmetic. The optimizer assumes U256 values fit in `usize` in `taint_range`.

Git commit: `ed95ffbdfe9e70a800fd0e227c0529940bbc2637`; LLVM 22.1.5; `resolc` v1.4.0+commit.ed95ffb.

Suggested fix: use checked arithmetic (`.checked_add()`, `usize::try_from`) in `taint_range` at `heap_opt.rs:724` rather than assuming U256 values fit in `usize`.

Several aliases share the same panic site: large MStore offsets and zero-length copy paths all trigger the same "attempt to add with overflow" at `heap_opt.rs:724`.

PR #582, opened 2026-07-28
Fix integer overflow in heap_opt range walk on top-of-memory access (…

kvpanch:
…#581)

The word-stepping loops in `taint_range`, `mark_escaping_range`, and `mark_escaping_and_tainted_range` walked a static range with `while word < end { word += 32 }`. When an access sits near the top of the address space, `end = address.saturating_add(size)` saturates at `u64::MAX`, and the final `word += 32` steps past `u64::MAX`, panicking with "attempt to add with overflow".

Under `--newyork` this is reachable from valid Yul: folding `add(mul(MAX_U256, 1), 0x41)` to `0x40` lets an mstore corrupt the free-memory-pointer word with `u64::MAX`; mem_opt then forwards `mload(0x40)` to that literal, turning a later `mstore(fmp, _)` into a static unaligned store at `u64::MAX` that taints its covered words.

Iterate the already-bounded `num_words` count with a saturating step so the walk covers the same words without overflowing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- `--newyork`: an ICE ("attempt to add with overflow") while analyzing a static memory access at the top of the address space, reachable from valid Yul where an `mstore` corrupts the free-memory-pointer word and a later store is forwarded to `u64::MAX`. [#582](https://github.com/paritytech/revive/pull/582)
+
 ## v1.4.0
 
 Supported `polkadot-sdk` rev: `2604.2.0`
--- a/crates/integration/src/tests.rs
+++ b/crates/integration/src/tests.rs
@@ -4908,6 +4908,25 @@ fn msize_observes_dead_store_expansion() {
     run_differential(actions);
 }
 
+/// ICE regression (newyork heap analysis): a static memory access at the top of the address space
+/// must not overflow the word walk. `add(mul(MAX_U256, 1), 0x41)` folds to `0x40`, corrupting the
+/// free-memory-pointer word with `u64::MAX`; `mem_opt` then forwards a store to that literal and
+/// the walk panicked with "attempt to add with overflow". Compiling is the regression; the call
+/// pins runtime behaviour against solc-EVM.
+#[test]
+fn heap_range_walk_at_top_of_memory() {
+    let mut actions = instantiate_yul("contracts/HeapRangeOverflowBug.yul", "HeapRangeOverflowBug");
+    actions.push(Call {
+        origin: TestAddress::Alice,
+        dest: TestAddress::Instantiated(0),
+        value: 0,
+        gas_limit: Some(GAS_LIMIT),
+        storage_deposit_limit: None,
+        data: vec![],
+    });
+    run_differential(actions);
+}
+
 /// Regression (newyork FMP range proof): a `calldatacopy` whose *dynamic*
 /// destination can land on the free-memory-pointer word `[0x40, 0x60)` corrupts
 /// the FMP, but only *static* copy destinations flagged `fmp_could_be_unbounded`.
--- a/crates/newyork/src/heap_opt.rs
+++ b/crates/newyork/src/heap_opt.rs
@@ -31,6 +31,32 @@ use revive_common::BYTE_LENGTH_WORD;
 /// this is treated as a dynamic escape instead.
 const MAX_RANGE_WORDS: u64 = 4096;
 
+/// The word-aligned words that `[address, address + size)` covers, start rounded down so a
+/// partial leading word is included: 32 bytes at `0x30` yields `0x20` and `0x40`.
+///
+/// [`None`] when the range exceeds [`MAX_RANGE_WORDS`]; callers then record only its first word
+/// and set their dynamic flag.
+///
+/// Stepping a bounded count saturatingly, rather than walking to a static end, is what keeps an
+/// access at the top of memory from overflowing `u64` — reachable from valid Yul via a corrupted
+/// free-memory pointer.
+fn range_words(address: u64, size: u64) -> Option<impl Iterator<Item = u64>> {
+    let first_word = word_align(address);
+    let num_words = address
+        .saturating_add(size)
+        .saturating_sub(first_word)
+        .saturating_add(BYTE_LENGTH_WORD as u64 - 1)
+        / BYTE_LENGTH_WORD as u64;
+
+    (num_words <= MAX_RANGE_WORDS).then(|| {
+        (0..num_words).scan(first_word, |word, _| {
+            let current = *word;
+            *word = word.saturating_add(BYTE_LENGTH_WORD as u64);
+            Some(current)
+        })
+    })
+}
+
 /// Classification of a memory access pattern.
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub enum AccessPattern {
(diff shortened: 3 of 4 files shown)


crates/newyork/src/heap_opt.rs
```diff
@@ -568,14 +568,15 @@ impl HeapAnalysis {
                 let range = end.saturating_sub(first_word);
                 let num_words =
                     range.saturating_add(BYTE_LENGTH_WORD as u64 - 1) / BYTE_LENGTH_WORD as u64;
+                // Records each covered word; same range walk and corner cases as `taint_range`.
```
xermicus, line 571: Please remove this and make a doc-comment instead

crates/newyork/src/heap_opt.rs
```diff
@@ -714,14 +715,26 @@ impl HeapAnalysis {
                     .saturating_sub(first_word)
                     .saturating_add(BYTE_LENGTH_WORD as u64 - 1)
                     / BYTE_LENGTH_WORD as u64;
+                // `[address, address + size)` spans `num_words` 32-byte words starting at
```
xermicus, line 718: Same here, please remove and make a proper doc comment instead.

xermicus (commented):
Could you please add the `MRE` from the bug report as integration test?

crates/newyork/src/heap_opt.rs
```diff
@@ -1043,16 +1056,17 @@ impl HeapAnalysis {
...
                     self.tainted_regions.insert(first_word);
                     self.has_dynamic_escapes = true;
                 } else {
                     let mut word = first_word;
-                    while word < end {
+                    for _ in 0..num_words {
                         self.escaping_regions.insert(word);
                         self.tainted_regions.insert(word);
-                        word += BYTE_LENGTH_WORD as u64;
+                        word = word.saturating_add(BYTE_LENGTH_WORD as u64);
```
elle-j, line 1069: From seeing how similar the updated sections are, we could consider adding e.g. a `fn range_words()` helper to extract the identical logic (`let end = ..`, `let first_word = ..`, `let num_words = ..`, `if num_words > MAX_RANGE_WORDS { ..`, etc.), having it return the range of words (an iterator over the words).

Other than easier maintenance and the stepping being handled in one place, I think the small differences at each call site would also be a bit clearer then (e.g. how they all handle `if num_words > MAX_RANGE_WORDS` being `true` or `false` a bit differently. The helper could return `None` when the range is wider than the max for instance). Let me know what you think.

crates/newyork/src/heap_opt.rs
elle-j, line 571: If by the "corner cases" you mean e.g. the `size == 0` case then at all locations they do handle this somewhat differently, so I'd drop the "same" wording here.

crates/newyork/src/heap_opt.rs
```diff
@@ -714,14 +715,26 @@ impl HeapAnalysis {
...
+                // word is included). Record each covered word.
+                //
+                // Corner cases:
+                // - Huge ranges (`num_words > MAX_RANGE_WORDS`) would iterate too long, so they
+                //   collapse to the first word plus a dynamic-access flag instead.
+                // - An access near the top of memory saturates `end` at `u64::MAX` (e.g. an
+                //   unaligned store through a corrupted free-memory pointer). Stepping the
+                //   bounded `num_words` count with a saturating add keeps the final increment
+                //   from overflowing `u64` — a plain `word += 32` panics here.
+                // - `size == 0` is excluded by the match guard (an empty range taints nothing).
```
elle-j, line 729: This last comment line needs an update. The [(Some(address), _)](https://github.com/paritytech/revive/blob/a958f055ea3af0ac19a25bc4072872ce39c7dba7/crates/newyork/src/heap_opt.rs#L741-L744) arm adds a tainted region for `size == 0`.

crates/newyork/src/heap_opt.rs
```diff
@@ -2109,4 +2123,49 @@ mod tests {
             "post-dedup native mode must be disabled for the now-variable offset word"
         );
     }
+
+    /// `taint_range` taints exactly the word-aligned words a static range covers,
+    /// including the partial leading word when the start is unaligned.
+    #[test]
+    fn taint_range_covers_spanned_words() {
```
elle-j, line 2130: Currently only the `taint_range()` is tested. Could consider adding similar ones for the other updates.

elle-j (commented):
Could you add a changelog entry as well, referencing this PR? 👍

crates/integration/src/tests.rs
```diff
@@ -4908,6 +4908,25 @@ fn msize_observes_dead_store_expansion() {
     run_differential(actions);
 }
 
+/// ICE regression (newyork heap analysis): a static memory access at the top of the address space
+/// must not overflow the word walk. `add(mul(MAX_U256, 1), 0x41)` folds to `0x40`, corrupting the
+/// free-memory-pointer word with `u64::MAX`; `mem_opt` then forwards a store to that literal and
+/// the walk panicked with "attempt to add with overflow". Compiling is the regression; the call
+/// pins runtime behaviour against solc-EVM.
```
elle-j, line 4915: Nit:
```suggestion
/// pins runtime behavior against solc-EVM.
```

crates/newyork/src/heap_opt.rs
```diff
@@ -2109,4 +2110,159 @@ mod tests {
...
+        let mut analysis = HeapAnalysis::new();
+        let huge = (MAX_RANGE_WORDS + 1) * BYTE_LENGTH_WORD as u64;
+        analysis.taint_range(Some(0), Some(huge));
+        assert_eq!(analysis.tainted_regions, BTreeSet::from([0]));
+        assert!(analysis.has_dynamic_accesses);
+    }
+
+    /// An access at the top of the address space must not overflow the word walk.
+    #[test]
+    fn taint_range_top_of_memory_does_not_overflow() {
```
elle-j, line 2146: We could add an assertion on the `analysis.has_dynamic_access` in this test as well.

crates/newyork/src/heap_opt.rs
```diff
@@ -2109,4 +2110,159 @@ mod tests {
...
+            (MAX_RANGE_WORDS + 1) * BYTE_LENGTH_WORD as u64,
+        );
+        analysis.mark_escaping_range(&offset, &length);
+        assert_eq!(analysis.escaping_regions, BTreeSet::from([0]));
+        assert!(analysis.has_dynamic_escapes);
+    }
+
+    /// The overflowing range reaches this caller through the same helper.
+    #[test]
+    fn mark_escaping_range_top_of_memory_does_not_overflow() {
```
elle-j, line 2204: We could add an assertion on the `analysis.has_dynamic_escapes` in this test as well.

crates/newyork/src/heap_opt.rs
```diff
@@ -2109,4 +2110,159 @@ mod tests {
...
+        let length = offset_value(&mut analysis, 2, BYTE_LENGTH_WORD as u64);
+        analysis.mark_escaping_and_tainted_range(&offset, &length);
+        assert_eq!(analysis.escaping_regions, BTreeSet::from([0x20, 0x40]));
+        assert_eq!(analysis.tainted_regions, BTreeSet::from([0x20, 0x40]));
+        assert!(!analysis.has_dynamic_escapes);
+    }
+
+    /// The overflowing range through the both-sets walk.
+    #[test]
+    fn mark_escaping_and_tainted_range_top_of_memory_does_not_overflow() {
```
elle-j, line 2229: We could add an assertion on the `analysis.has_dynamic_escapes` in this test as well.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
```
elle-j, line 4:
```suggestion
## Unreleased

Supported `polkadot-sdk` rev: `2604.2.0`

```

elle-j (approved):
Nice, thanks!

PR #594, opened 2026-08-14
Enable the wide integer extension (XReviveVec)

kvpanch:
EVM words are 256 bits and RISC-V registers 64, so every wide operation becomes a four-limb chain. With no carry flag each carry costs an `sltu` to produce and another to consume, and anything wider than two registers is passed by reference: 56 bytes for one `add i256`, and a 192-byte frame to call with three of them.

Request `+xrevivevec`, which makes i256 a machine type held in VRM2 -- one instruction per wide operation, arguments in registers. -30% code over the 15 benchmark contracts.

Experimental: the PolkaVM linker cannot decode the custom-2 encodings, so on this branch resolc fails at link for every contract.


xermicus (commented):
Thats a good first indicator thanks! Small nit to start with: Can we name the extension somelike XRreviveWide, the `Vec` in XReviveVec confused me slightly (you'd think it's a custom vector extension which isn't very accurate I think)

PR #597, opened 2026-08-23
yul: Compile objects that carry deploy code only

### LINKED ISSUE #490: resolc fails on deploy-only Yul objects
`resolc --yul` fails when a Yul object has no `_deployed` sub-object, while `solc` compiles it fine.

```yul
object "Test" {
    code {
        { mstore(0, 42) return(0, 32) }
    }
}
```

`__runtime` is created unconditionally even when no runtime code exists, leaving its entry block without a terminator.

dimartiro:
# Description

Closes #490

`resolc --yul` rejected any Yul object that carries deploy code only, that is one without a `_deployed` runtime sub-object, while `solc` compiles it:

```yul
object "Test" {
    code {
        { mstore(0, 42) return(0, 32) }
    }
}
```

```
Error: The contract `deploy_only.yul:Test` unoptimized LLVM IR verification error:
Basic Block in function '__runtime' does not have terminator!
label %entry
```

## Cause

`Object::declare` declares `PolkaVMRuntimeCodeFunction` unconditionally, so `__runtime` always exists with an empty `entry` block. `Object::into_llvm` only ever *defines* it from the `_deployed` branch, which is reached through `if let Some(object) = self.inner_object`. With no sub-object the function stays declared and undefined, and the module fails verification.

Not declaring it is not an option: `Entry::leave_entry` looks `__runtime` up and bails with `Contract runtime code not found`, since `__entry` dispatches to either `__deploy` or `__runtime` on the call flag.

## Fix

`Object::parse` now materializes the runtime sub-object when the source omits it, so the AST that reaches code generation always has one.

Doing it in the parser rather than special casing the absence in code generation keeps every consumer on the path it already handles. That matters here because both IR pipelines read this field: `newyork`'s `YulTranslator::translate_object` has the same `if let Some(inner_object)` shape, and `--newyork` reproduces the identical error. One parser-level change fixes both.

The synthesized body is an explicit `stop()`. `Code::into_llvm` terminates every code block with an implicit `stop` anyway ("The EVM lets the code return implicitly"), so this is exactly what an empty runtime block lowers to, and it matches the EVM behaviour of calling an account with no code: success with empty return data.

I did not use an empty block, for two reasons. It would leave `__runtime` as `entry: br %return` / `return: unreachable`, and that `unreachable` is now *reachable* rather than a dead tail, which is UB the optimizer is free to exploit — `__runtime` is `Private` and called only from `__entry`, so folding it away could turn a call into a constructor run. And separately, an empty runtime block currently trips an assertion in `polkavm-linker` under `--newyork` (see below), which the `stop()` avoids.

Tests added:
- `revive-yul`: the deploy-only object gets the implicit runtime object with a `stop()` body, and an explicit `_deployed` object is not clobbered by it.
- `resolc`: `compiles_object_without_a_runtime_sub_object` compiles the new `deploy_only_object.yul` fixture end to end and cross-checks that `solc` agrees on the exit code, which is the premise of the issue.

### DIFF
--- a/crates/resolc/src/tests/cli/yul.rs
+++ b/crates/resolc/src/tests/cli/yul.rs
@@ -2,7 +2,7 @@
 
 use crate::cli_utils::{
     assert_command_failure, assert_command_success, assert_equal_exit_codes, execute_resolc,
-    execute_solc, RESOLC_YUL_FLAG, SOLC_YUL_FLAG, YUL_CONTRACT_PATH,
+    execute_solc, RESOLC_YUL_FLAG, SOLC_YUL_FLAG, YUL_CONTRACT_PATH, YUL_DEPLOY_ONLY_OBJECT_PATH,
     YUL_DUPLICATE_FUNCTIONS_DEEP_NESTING_PATH, YUL_DUPLICATE_FUNCTIONS_SWITCH_PATH,
     YUL_INVALID_HEX_NIBBLES_PATH,
 };
@@ -62,4 +62,17 @@ fn duplicate_functions_deep_nesting() {
         &resolc_result,
         "Duplicate function names in deeply nested switch cases",
     );
+}
+
+#[test]
+fn compiles_object_without_a_runtime_sub_object() {
+    let resolc_result = execute_resolc(&[YUL_DEPLOY_ONLY_OBJECT_PATH, RESOLC_YUL_FLAG, "--bin"]);
+    assert_command_success(
+        &resolc_result,
+        "An object carrying deploy code only, without a `_deployed` sub-object",
+    );
+
+    // `solc` accepts this too, which is the whole point: the two front-ends should agree.
+    let solc_result = execute_solc(&[YUL_DEPLOY_ONLY_OBJECT_PATH, SOLC_YUL_FLAG]);
+    assert_equal_exit_codes(&solc_result, &resolc_result);
 }
--- a/crates/yul/src/parser/statement/object.rs
+++ b/crates/yul/src/parser/statement/object.rs
@@ -20,7 +20,12 @@ use crate::lexer::token::location::Location;
 use crate::lexer::token::Token;
 use crate::lexer::Lexer;
 use crate::parser::error::Error as ParserError;
+use crate::parser::statement::block::Block;
 use crate::parser::statement::code::Code;
+use crate::parser::statement::expression::function_call::name::Name;
+use crate::parser::statement::expression::function_call::FunctionCall;
+use crate::parser::statement::expression::Expression;
+use crate::parser::statement::Statement;
 use crate::visitor::AstNode;
 use crate::visitor::AstVisitor;
 
@@ -117,7 +122,7 @@ impl Object {
                     factory_dependencies.extend(object.factory_dependencies.drain());
                     Some(Box::new(object))
                 }
-                _ => None,
+                _ => Some(Box::new(Self::implicit_runtime_code(&identifier, location))),
             };
 
             if let Token {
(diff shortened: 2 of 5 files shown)


crates/resolc/src/tests/cli/yul.rs
```diff
@@ -63,3 +63,16 @@ fn duplicate_functions_deep_nesting() {
...
+
+#[test]
+fn compiles_object_without_a_runtime_sub_object() {
+    let resolc_result = execute_resolc(&[YUL_DEPLOY_ONLY_OBJECT_PATH, RESOLC_YUL_FLAG, "--bin"]);
+    assert_command_success(
+        &resolc_result,
+        "An object carrying deploy code only, without a `_deployed` sub-object",
+    );
+
+    // `solc` accepts this too, which is the whole point: the two front-ends should agree.
```
xermicus, line 75:
```suggestion
```

This is the default behavior so no comment needed.

crates/yul/src/parser/statement/object.rs
```diff
@@ -488,4 +533,71 @@ object "Test" {
...
+        let object = Object::parse(&mut lexer, None).expect("the object should parse");
+
+        let inner = object
+            .inner_object
+            .expect("a deploy-only object should get an implicit runtime object");
+        assert_eq!(inner.identifier, "Test_deployed");
+        assert_eq!(inner.inner_object, None);
+
+        // An explicit `stop`, so that `__runtime` is emitted with a terminator.
+        assert_eq!(inner.code.block.statements.len(), 1);
```
xermicus, line 556:
```suggestion
        assert_eq!(inner.code.block.statements.len(), 1, "expected one statement");
```

xermicus (approved):
lgtm thanks

crates/yul/src/parser/statement/object.rs
```diff
@@ -173,6 +178,43 @@ impl Object {
...
+    /// return with empty data, which is exactly what an empty Yul code block lowers to (see
+    /// `Code::into_llvm`, which terminates every code block with an implicit `stop`).
+    ///
+    /// Materializing the sub-object here instead of special casing its absence in code
+    /// generation keeps every consumer of the AST -- both IR pipelines, the visitors and the
+    /// analyses -- on the one path they already handle.
+    ///
+    /// The body is an explicit `stop` rather than an empty block. The two are equivalent, but an
+    /// empty runtime block currently trips an assertion in `polkavm-linker` when lowered through
+    /// the `--newyork` pipeline.
```
xermicus, line 195:
```suggestion
    /// A short-hand constructor returning an object with a call to `stop`.
```

crates/resolc/src/tests/cli/yul.rs
dimartiro: Removed, thanks — applied in 22cd30b.

crates/yul/src/parser/statement/object.rs
dimartiro: Applied. One deviation worth flagging: `rustfmt` does not keep it on a single line, so it landed as

```rust
assert_eq!(
    inner.code.block.statements.len(),
    1,
    "expected one statement"
);
```

The one-line form fails `cargo fmt --check`.

crates/yul/src/parser/statement/object.rs
dimartiro: Applied — the rationale that was in the doc comment lives in the PR description instead.

xermicus:
@dimartiro this is good to go merge conflicts are gone

dimartiro:
@xermicus conflicts solved

PR #598, opened 2026-08-23
newyork: Terminate object code with the implicit EVM return

dimartiro:
# Description

`--newyork` crashes in the linker on a Yul object whose `_deployed` runtime code block is empty:

```yul
object "T" {
    code { stop() }
    object "T_deployed" {
        code { }
    }
}
```

```
thread 'main' panicked at polkavm-linker-0.35.0/src/program_from_elf.rs:10396:
internal error: inconsistent reachability after optimization; this is a bug, please report it!
```

The default pipeline compiles the same input fine.

## Cause

The panic is fallout. The actual problem is a silent miscompile one step earlier.

`Code::into_llvm` in the legacy pipeline terminates every code block with an implicit `stop`
("The EVM lets the code return implicitly"). `LlvmCodegen::generate_object` in `newyork` does not,
so a block that falls off the end reaches the

```llvm
define private void @__runtime() {
entry:
  br label %return
return:
  ret void        ; <-- __runtime returns
}
```

`__entry` is marked `NoReturn` and puts `unreatwo calls. A `__runtime`
that returns therefore makes the `is_deploy ==d LLVM removes the
dispatch entirely:

```llvm
define void @__entry(i1 %0) {
entry:
  tail call fastcc void @__deploy()
  unreachable
}
```

Calling the deployed contract would run the **constructor**. The linker then sees exports whose
reachability no longer agrees and asserts.

## Fix

Append the implicit EVM return in `generate_obcy pipeline does. The
runtime function then ends in the exit syscall and keeps its `noreturn` shape, so the dispatch
edge stays live:

```llvm
define private void @__runtime() {
entry:
  call void @__revive_exit(i32 0, i256 0, i256 0)
  br label %return
return:
  ret void        ; now unreachable, cleaned up by LLVM
}
```

```llvm
define void @__entry(i1 %0) {
entry:
  br i1 %0, label %deploy_code_call_block, label %runtime_code_call_block
  ...
}
```

## Verification

| input | default | `--newyork` before | `--newyork` after |
| --- | --- | --- | --- |
| `_deployed` with an empty block | ✅ | 💥 linker assert | ✅ |
| `_deployed` with `stop()` | ✅ | ✅ | ✅ |

Across `crates/integration/contracts/*.sol`:

- **default pipeline: 83/83 byte-identical**, as expected since this only touches `newyork`.
- **`--newyork`: 82/83 byte-identical.** The ostruct.sol`, and it changes
  in the right direction. `selfdestruct` lowers to a `terminate` call that neither terminates the
  basic block nor is marked `noreturn`, so itsfall through and return
  into `__entry`'s `unreachable`. It now ends peline already ends it:

  ```llvm
    call void @terminate(...)
    call void @seal_return(i32 0, i32 ptrtoint2), i32 0)
    unreachable
  ```

  So the diff is `newyork` converging onto the which is the point of the
  change.

Test added: `compiles_newyork_empty_runtime_obpty_runtime_object.yul`
fixture through `--yul --newyork --bin`, mirroring the existing
`compiles_newyork_folded_guard_inlined_loop` r

`cargo test -p revive-newyork` 92 passed, `cargo test -p resolc --lib` 69 passed, clippy clean.

## Note

Found while working on #490, but independent of it: #490 is an object with **no** `_deployed` sub-object at all, which fails in both pipelines with an LLVM verifier error. This one is an object **with** a `_deployed` sub-object whoseaffects `--newyork`. Neither fix depends on the other.

### DIFF
--- a/crates/newyork/src/to_llvm.rs
+++ b/crates/newyork/src/to_llvm.rs
@@ -3496,6 +3496,14 @@ impl<'ctx> LlvmCodegen<'ctx> {
 
         self.generate_block(&object.code, context)?;
 
+        // The EVM lets the code return implicitly, so terminate the block the way the legacy
+        // pipeline does in `Code::into_llvm`. Without it a code block that falls off the end
+        // reaches the function's `ret void`, which contradicts `__entry` being `noreturn`: LLVM
+        // then proves the dispatch edge into this function unreachable and folds it away, so
+        // calling the deployed contract would run the constructor.
+        revive_llvm_context::polkavm_evm_return::stop(context)
+            .map_err(|error| CodegenError::Llvm(error.to_string()))?;
+
         context
             .set_debug_location(0, 0, None)
             .map_err(|error| CodegenError::Llvm(error.to_string()))?;
--- a/crates/resolc/src/tests/cli/bin.rs
+++ b/crates/resolc/src/tests/cli/bin.rs
@@ -5,7 +5,7 @@ use crate::{
         absolute_path, assert_command_success, execute_command, execute_resolc, CommandResult,
         ResolcOptSettings, SolcOptSettings, SOLIDITY_CONTRACT_PATH,
         SOLIDITY_FOLDED_GUARD_INLINED_LOOP_PATH, STANDARD_JSON_CONTRACTS_PATH,
-        YUL_MEMSET_CONTRACT_PATH,
+        YUL_EMPTY_RUNTIME_OBJECT_PATH, YUL_MEMSET_CONTRACT_PATH,
     },
     SolcCompiler,
 };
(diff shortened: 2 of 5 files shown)


crates/newyork/src/to_llvm.rs
```diff
@@ -3496,6 +3496,14 @@ impl<'ctx> LlvmCodegen<'ctx> {
 
         self.generate_block(&object.code, context)?;
 
+        // The EVM lets the code return implicitly, so terminate the block the way the legacy
+        // pipeline does in `Code::into_llvm`. Without it a code block that falls off the end
+        // reaches the function's `ret void`, which contradicts `__entry` being `noreturn`: LLVM
+        // then proves the dispatch edge into this function unreachable and folds it away, so
+        // calling the deployed contract would run the constructor.
```
xermicus, line 3503:
```suggestion
        // The EVM lets the code return implicitly.
```

crates/resolc/src/tests/cli/bin.rs
```diff
@@ -180,6 +180,20 @@ fn compiles_newyork_folded_guard_inlined_loop() {
     assert_pvm_blob(&result);
 }
 
+/// Compile regression: a `_deployed` object with an empty code block crashed the newyork
+/// pipeline in `polkavm-linker` with "inconsistent reachability after optimization". The code
+/// block fell through to the function's `ret void`, which contradicts `__entry` being
+/// `noreturn`, so LLVM folded the runtime dispatch edge away and a call would have run the
+/// constructor. The implicit `stop` the legacy pipeline appends keeps the edge live.
```
xermicus, line 187:
```suggestion
```

The name of the test is already self-describing and unambiguous to me.

xermicus (approved):
lgtm!

crates/newyork/src/to_llvm.rs
dimartiro: Trimmed to the one-liner — the reasoning about the folded dispatch edge is in the PR description.

crates/resolc/src/tests/cli/bin.rs
dimartiro: Agreed, removed — the test name carries it.

crates/newyork/src/to_llvm.rs
xermicus: This didn't get through though

dimartiro: Pushed in e87562a — the commit had been sitting unpushed locally, sorry for the noise. It's the one-liner now.

crates/resolc/src/tests/cli/bin.rs
dimartiro: Also in e87562a — same unpushed commit, so this one hadn't landed either. The doc comment is gone; the test name carries it.

PR #599, opened 2026-08-23
resolc: Detect undeployable contracts at link time

### LINKED ISSUE #96: Detect problematic blob size at link time
Emit a compiler warning (or error) if we detect the final blob size to exceed the pallets [limits](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/revive/src/limits.rs).

### Tasks
- [ ] Add a command line flag to provide the memory limit (different chains can exhibit different limits), and assume AH as default
- [ ] Helper to calculate the maximum linked blob size given the static memory buffer and the the pallet limits
- [ ] Check it after linting and emit a warning if over the threshold
- [ ] By default it should error out but we want to have a flag to ignore it

dimartiro:
# Description

Closes #96

A contract whose linked blob exceeds the target runtime's limits compiles fine and is then rejected by `pallet_revive` on deployment. This reproduces the pallet's checks at link time so the problem surfaces where it can still be acted on.

## What it checks

`pallet_revive::limits::code::enforce` rejects a blob on four grounds the compiler can evaluate: the blob length (`BLOB_BYTES`), the longest basic block (`BASIC_BLOCK_SIZE`), and the purgeable and baseline interpreter memory (`PURGABLE_MEMORY_LIMIT`, `BASELINE_MEMORY_LIMIT`). All four are now checked, and each reports the measured value against the limit it broke.

The memory figures are not approximated. The pallet derives them from `ProgramBlob::estimate_interpreter_memory_usage`, and `revive` already depends on `polkavm-common`, so `crates/linker/src/limits.rs` calls the same API with the same arguments and reaches the same verdict rather than a conservative guess.

## Interface

- `--memory-limit <BYTES>`: the target runtime's baseline interpreter memory budget, defaulting to Asset Hub's. The compiler cannot know which runtime a blob is destined for, and the same chain at a different runtime version can carry a different budget.
- `--ignore-memory-limit`: report violations as warnings and emit the contract anyway.

Exceeding the budget is an error by default, on the grounds that a contract which cannot be deployed is not a useful build artifact.

## Why this matters in practice

The baseline budget is where it bites, and it is easy to cross without touching a line of Solidity, because the static heap and stack buffers are part of it. Compiling `Baseline.sol` unchanged:

| `--heap-size` | baseline memory | verdict |
| --- | --- | --- |
| 64 KiB | 201,542 | ok |
| 256 KiB | 398,150 | ok |
| 1 MiB | 1,184,582 | ok |
| 2 MiB | **2,233,159** | over the 1,572,864 budget |

Before this change the last row compiled successfully and produced a contract that could never be deployed. It now fails with the numbers and a pointer at the knob to turn:

```
Error: Baseline.sol:Baseline cannot be deployed to the target runtime: the contract needs 2233159 bytes of baseline interpreter memory but the runtime provides at most 1572864 bytes; the static heap and stack buffers are part of this budget, so lowering `--heap-size` or `--stack-size` frees room for code. Pass `--ignore-memory-limit` to compile anyway, or `--memory-limit` if the target runtime has a different budget.
```


crates/resolc/src/build/mod.rs
```diff
@@ -86,6 +88,12 @@ impl Build {
                     !debug_config.emit_debug_info,
                 ) {
                     Ok((memory_buffer_linked, ObjectFormat::PVM)) => {
+                        self.messages.extend(check_deployment_limits(
```
xermicus, line 91: `Build::write_to_standard_json` considers errors, so does this doesn't work in std json mode? Would be nice to have a test for that too.

xermicus (commented):
Thanks for the PR!

crates/resolc/src/build/mod.rs
dimartiro: Good catch — it didn't. `write_to_standard_json` only forwarded errors from `results` and dropped `messages`, which is where the deployment-limit diagnostics (and other link-time issues) live.

They now land in the standard JSON `errors` array. Added coverage for `--standard-json`: the error, `--ignore-memory-limit` as a warning, and a larger `--memory-limit`.

Addressed in 98e6232.

crates/linker/src/limits.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub purgable_memory_limit: u32,
+    /// `limits::code::BASELINE_MEMORY_LIMIT`: memory held for the contract's whole life time.
+    pub baseline_memory_limit: u32,
+    /// `limits::PAGE_SIZE`.
+    pub page_size: u32,
+}
+
+impl PalletLimits {
+    /// The limits of the Asset Hub runtime.
+    pub const ASSET_HUB: Self = Self {
```
xermicus, line 40: Instead of duplicating code here, we should re-use this from pallet-revive directly. It's already in the dependency tree.

xermicus (changes requested):

crates/linker/src/limits.rs
dimartiro: I'd like to, but `limits` isn't reachable from the published crate

Two ways forward, happy with either:

1. Keep the constants, and I'll open a polkadot-sdk PR making `limits` public so a later bump can replace them with the real thing.
2. Do the polkadot-sdk PR first and block this one on the release, if you'd rather not merge the constants at all.

xermicus: I see, thank you. In that case I'd prefer to do 2. and wait until it's available

dimartiro: Done [here](https://github.com/paritytech/polkadot-sdk/pull/13300) just need someone to approve it :D

xermicus: Good stuff thank you! Note that we will also need to wait for a polkadot-sdk release once this is in, I'm afraid you have to wait a little

PR #608, opened 2026-09-14
support solc v0.8.37

xermicus:
- Support solc v0.8.37
- The new `SLOTNUM` builtin is not supported on pallet-revive


crates/newyork/src/from_yul.rs
```diff
@@ -646,6 +646,7 @@ impl YulTranslator {
             FunctionName::Difficulty | FunctionName::Prevrandao => Ok(Expression::Difficulty),
             FunctionName::BaseFee => Ok(Expression::BaseFee),
             FunctionName::BlobBaseFee => Ok(Expression::BlobBaseFee),
+            FunctionName::SlotNum => Err(TranslationError::Unsupported("slotnum".to_string())),
```
elle-j, line 649: Nit: Any reason it's placed here rather than a match arm lower down? We could keep unsupported ones at the end to hit those cases last.

elle-j (approved):
Read thru the release/blog post as well and LGTM 👍

crates/newyork/src/from_yul.rs
xermicus: Thanks for the suggestion. It was put there because its numerically the next opcode after `BLOBBASEFEE`. I think we'll leave it like that for now (the unsupported ones could go all to the bottom but it's not ordered like this anyways).

